PR #5, opened 2015-09-01
Account for the recently added `read_exact` method to `Read`.

marcusklaas:
This makes zip-rs only compile on nightly rustc, which is a major downside.

How did `read_exact` previously work? Is this not a breaking change? I'm confused.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 //! A basic ZipReader/Writer crate
 
 #![warn(missing_docs)]
+#![feature(read_exact)]
 
 extern crate bzip2;
 extern crate flate2;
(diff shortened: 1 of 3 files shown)


mvdnes:
Thanks for bringing this to my attention. I am confused as well.

I am not merging this because working on stable is more important to me than working on nightly. There are some options I can think of to get them to work on both, but have not decided yet which is the best one.

marcusklaas:
I understand that working on stable is more important. This seems like a big compatibility break from the rustc team. Was hoping you had more insight.

marcusklaas:
Aaah, it's an aliasing issue! `read_exact` is now used from your other crate: http://mvdnes.github.io/rust-docs/podio/podio/index.html

mvdnes:
I have pushed a similar fix to the `nightly` branch (c97d6ed7) which works on both 1.2 stable and 1.4 nightly. I will probably merge this into master, but first want to do a little more investigation on the breaking changes policy of rustc.

marcusklaas:
Looks good! This was the proposed solution when I asked about the issue on the rust IRC too. I think aliasing breaks like this are considered acceptable by the rust team.

PR #7, opened 2015-10-05
Remove `no-bzip2` feature because it doesn't work as I expected

alexbool:
An "oops" PR :)
I realized that `no-bzip2` does not do anything at all, and the right way to get rid of `bzip2` is just saying `zip = { version = "*", default-features = false }` in `Cargo.toml`.


mvdnes:
Thanks for fix!

PR #12, opened 2016-04-21
Support for external file attributes

anti-social:


mvdnes:
Thanks for the suggestion! I really think we should implement this.

I would like to structure the `ZipFileData` differently however by:
* exposing the `system` attribute as an enum rather than a `u8`
* not exposing the `external_attributes` directly
* I am unsure what `mode` should be. Please say if you think otherwise, but I would rename it to `unix_mode` or `unix_permission`, and let it return an `Option<u32>`, so that it won't be set to 0 mistakenly.

One last note: there is a bug in the octal numbers. In rust you write them like `0o644`, so with a zero and small leter O as prefix.

anti-social:
My first idea was to make `system` attribute as enum. But there is 20 variants of the enum and most of them I think are useless.

mvdnes:
Thank you for these additions.
Regarding system: we can either just create `Unix/Dos/Unknown` or perhaps system should then also remain private (for now)

I will merge this later/tomorrow

anti-social:
I also made `version` private because probably it would be nice to split `version` into (major, minor) tuple.

mvdnes:
Thank you very much! :+1:

PR #13, opened 2016-04-29
Add crc32 accessor for ZipFile

efyang:
This adds an accessor for the crc32 hash of the ```ZipFile```. This is probably useful for those that want to check the integrity of the file.


mvdnes:
Thanks for the suggestion! :+1:

PR #19, opened 2016-10-04
Implement Display trait for CompressionMethod

rowedonalde:


mvdnes:
Thank you for the contribution.
Before I merge this, could you please explain why you would want to use the display trait instead of using the debug trait directly?

rowedonalde:
This patch is meant to provide expected behavior in user-facing release vs. programmer-facing debug contexts. Per [Debug's docs](https://doc.rust-lang.org/std/fmt/trait.Debug.html), "Debug should format the output in a programmer-facing, debugging context," whereas according to [Display's docs](https://doc.rust-lang.org/std/fmt/trait.Display.html), "Display is similar to Debug, but Display is for user-facing output." For example, if someone wanted to write a CLI utility that detailed the contents of an archive, they could use `{}` as expected for printing name, comment, size, etc., but they'd inconsistently have to use `{:?}` for printing the compression method.

mvdnes:
Ah yes, for CLI purposes this seems the best way to go. Thanks! :+1:

PR #21, opened 2016-10-25
set unix_mode and last_modified fields when creating a zipfile

ultimaRatio:
Hi,

Nothing fancy, just a quick hack, but it may help others.
unix_modes need to be represented in a better way, perhaps with an enum

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -164,6 +164,58 @@ impl<W: Write+io::Seek> ZipWriter<W>
 
         Ok(())
     }
+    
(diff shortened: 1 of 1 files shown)


mvdnes:
I like the idea, but do have some changes to the API. Will look at it soon!

mvdnes:
I have used some of your ideas, and pushed a new version in commit d2d19f6 that allow setting of the permissions. It changes the API, so you want to switch to version `0.2` of the crate.
Thanks for all the input!

PR #29, opened 2017-03-06
Fix writing to buffers.

stanmihai4:
Using SeekFrom::End(0) in finish_file for a buffer was movind the position
the the end of the buffer, not the last written byte.


mvdnes:
You are right! Great work noticing this subtle bug 👍

PR #31, opened 2017-03-20
Add function for parallel compression

vandenoever:
Sometimes one wants to compress a very large file or stream. With this function
it is possible to do this in parallel.

This requires crc32_combine which is contained only in libz_sys for now. Also it uses a function from flate2 that is not merged yet. Pending https://github.com/alexcrichton/flate2-rs/pull/65 I'm putting this PR up for consideration.


vandenoever:
The required function is now in flate2 master.

vandenoever:
When https://github.com/alexcrichton/flate2-rs/issues/64 is resolved, the direct dependency on libz_sys can go.

mvdnes:
Looks neat. Once the libz_sys dependency is gone or fixed for Windows, I will definitely take a look!

vandenoever:
I've opened https://github.com/alexcrichton/flate2-rs/pull/68 which exposes zlib and minizip crc code in Rust. If this is accepted in some form, zip-rs code and this pull request can be simplified.

jrmuizel:
alexcrichton/flate2-rs#68 has been merged. Can you post the simplified version?

vandenoever:
@jrmuizel done

Tests pass.

I've another branch that eschews Read and uses a StreamWriter trait instead. The difference is in how the caller passes the data. In one case a `Read` implementation should be supplied on which `read()` is called repeatedly. In the other case a `StreamWriter` is supplied tha can be called to have a function write all the data into a `Write`. That might be more convenient. For example some libraries have a function that writes out a data structure in one go into a `Write`. Of course it's possible to supply both methods.

```rust
pub trait Read {
    fn read(&mut self, buf: &mut [u8]) -> Result<usize>;
}
fn compress_substream<R>(stream: &mut R) -> ZipResult<(u64, u64, Vec<u8>)> where R: io::Read {...}
```

```rust
pub trait StreamWriter {
    /// Write data into a zip entry.
    /// This function consumes `self` so all data should be written in one call
    /// to `StreamWriter::write`.
    fn write(self, stream: &mut io::Write) -> io::Result<()>;
}
fn compress_substream<W>(writer: W) -> ZipResult<(u64, u64, Vec<u8>)> where W: StreamWriter + Send {...}
```

jrmuizel:
@mvdnes can you review this?

mvdnes:
* Some variable names are unclear. I do not understand all of the code, mostly due to things like `c.0, r.0, r.1`.
* Is the compressed in paralllel in memory, and then written to disk sequentially? If so, then I think this might cause more issues than what it is trying to solve.
* It would be better to put this behind a feature, as for some the rayon dependency might be an issue.

mvdnes:
Closing this due to inactivity. Feel free to comment if you still want to do something with this.

PR #34, opened 2017-04-08
Handle data prepended to the zip

sagebind:
The official spec for the Zip format allows arbitrary data to be prepended to the beginning of a Zip file, but doesn't give much help on what that does to the offsets stored in the file headers. This is something that most Zip libraries account for, and I actually have need of this crate supporting it  (gonna use it for self-extracting Zip files).

I looked first to `libarchive` on how to implement this. They handle it by [seeking through the file and looking for what appears to be the first central directory entry](https://github.com/libarchive/libarchive/blob/master/libarchive/archive_read_support_format_zip.c). This is actually unnecessary since we know what position we seek to when we find the start of the central directory end. Thus the offset of the entire archive can be calculated as:

> start of CDE - CD size - CD offset

This PR adds this additional check after reading the footer and before reading anything else, and adjusts all offsets by the above amount.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -54,6 +54,7 @@ pub struct ZipArchive<R: Read + io::Seek>
     reader: R,
     files: Vec<ZipFileData>,
     names_map: HashMap<String, usize>,
+    offset: u32,
 }
 
 enum ZipFileReader<'a> {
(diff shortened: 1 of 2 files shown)


mvdnes:
Thanks! This was indeed the intended behaviour.

PR #35, opened 2017-04-17
Add data_start property to ZipFile

bbqsrc:
Hey!

In a project I'm working on, I need to be able to access the `data_start` property of a file in a zip file. The use case is: the zip file is stored with no compression, and I am memory mapping each file directly. It works very well. :smile:

It would be greatly appreciated if you could accept this small pull request.


mvdnes:
Sure, seems like something some people could use. 👍

Issue #18, opened 2016-09-24
Can ZipArchive::by_index/by_name not be mut?


icefoxen:
It just feels like a weird abstraction break to have a fundamentally read-only operation require the ZipArchive to be mutable, when all that's ever changing is the internal reader object which won't be shared by anything else anyway.  It seems like wrapping it in a `RefCell` would let you pretend everything is immutable, since the API is entirely immutable anyway.

I'm not sure this is a good or feasible idea, it's just a suggestion.  I might try implementing it sometime and see how it works out.

mvdnes:
If you have a nice implementation I would be happy to take a look at it and merge it!
For the time being I have no real interest in implementing this myself, but would like it if this was fixed.

icefoxen:
Ok, I'll play around with it and see what I can come up with.

icefoxen:
Upon playing around, this might not be doable without changing how a `ZipFileReader` works somehow.  Even if you try to wrap the `ZipArchive`'s reader up in a `RefCell`, if you try to create two `ZipFileReader`'s you will end up with two mutable references both trying to screw with the underlying reader.  You really need an entirely separate reader for each `ZipFileReader`, each preserving where they are pointing in the file.  But you can't, generally, clone something implementing `Read`.

You can try faking it with a single reader by having each `ZipFileReader` remembering where its file offset is, seeking to that and updating it every time you read from the `ZipFileReader`, and maybe using unsafe to convince everything that this doesn't count as "mutation".  But I don't know about much the innards of this and it would basically break the current model which is to have a single `Read` type and just wrap that in layers of translation.

mvdnes:
Come to think of it, this seems closely related to issue #14. Feel free to continue the discussion there.

PR #36, opened 2017-04-18
Allow multiple access to ZipFileData

glasserc:
This is a provisory attempt at trying to address #18. The main use case I had is that I wanted to be able to look through a Zip file looking for one or two "magic" filenames. Because `by_name()` is `&mut`, it's impossible to look for a second file if the first one is absent. I also noticed the lack of an `Iterator` capability on any of the types in `zip::read`.

This PR separates reading `ZipFileData` out from reading `ZipFile`s. `by_name` and `by_index` now return `ZipFileData` values, and there is a new `open(&mut, &ZipFileData)` method that takes a `ZipFileData` and returns a `ZipFile`. Unfortunately, at first, a client can't call `open()` because the existence of a `ZipFileData` reference means that there's a borrow on the larger `ZipArchive`, preventing the `&mut` borrow we need for `open()`.

In order for this to work, we apply the Rust design pattern of [composing structs to achieve better borrowing](https://github.com/rust-unofficial/patterns/blob/master/patterns/compose-structs.md), namely by moving `open()` out of the same `struct` that has access to the `by_name`/`by_index` indices and into a new type. This type is a natural place to centralize access to the "reader" that we're using to read the `ZipFile` itself. Once you extract this functionality, the only thing left in the current `ZipArchive` type is its `by_name`/`by_index` indices, so rename `ZipArchive` to `ZipIndex` and create a new `ZipArchive` type that wraps the "reader". We can also add an `Iterator` type that walks the (new) `ZipArchive` and yields each entry in the directory. `ZipIndex`es can be built using this iterator; provide a convenience method `ZipArchive#index()` that allows this for convenience.

While I was here, I performed a few other cleanups:

- `rustfmt` everything
- Get rid of `ZipError::FileNotFound`; this error doesn't arise in any ZipArchive method, but only in `ZipIndex`, where it can be replaced with `Option`
- Move `unix_mode` from `ZipFile` to `ZipFileData`, because we needn't have opened a file to compute its UNIX mode.

I might have gotten a bit carried away here -- my main goal was to highlight the refactor of separating `ZipArchive` out from `ZipIndex` and offer it for consideration. I learned a lot about "thinking in Rust" in the process, so don't be afraid to reject this PR ;)

This PR makes almost no attempt to preserve backwards compatibility (it renames a few types), but because this library is only version 0.2.3, I didn't worry too much about it.

Feedback gratefully solicited :)


mvdnes:
This PR is a bit too complicated to merge at this point.
Sorry that it took a long time to have a verdict.

I am not really a fan of people running rustfmt on all the files, this touches too many things, and `git blame` will not give any reliable results anymore.

I think this can be solved in several ways:
* Run `by_name` in a new scope, so that the reference is dropped afterwards and the second file can be checked.
* We can add a `filenames` iterator to `ZipArchive`
* You can loop over all file indices, and check the names in the loop

Feel free to add to this discussion or open a new one, but I am closing the PR.

Issue #39, opened 2017-05-31
Does not support ZIP64


pkgw:
I just ran into a problem because I have a Zip file with ~80000 files in it, and zip-rs can't handle it since you need to support ZIP64 to work with archives with more than 65535 files.

This is pretty urgent for me so I am going to see what I can do to get my use case working. It seems like it would be good to have an issue open to track the general task, though, so here we are.

pkgw:
[Here's a branch](https://github.com/pkgw/zip-rs/tree/pr-basic-zip64) that fixes my proximate problem.

mvdnes:
This was one of the features I have been willing to add! Thanks for your initial work.
Unfortunately I do not have a lot of time at the moment to read the spec. I would greatly appreciate it if you could also add some tests.

Kixunil:
That branch works for me and solves my problem, so thank you very much! I'd like to help with further development but I'm low on time. :'(

Good luck!

mvdnes:
Closed in #44

PR #44, opened 2017-08-12
Very basic ZIP64 support

pkgw:
As per issue #39. This patch adds code that will locate and use a ZIP64 end-of-central-directory data structure if it is present in the file. This allows us to properly handle archives with more than 65535 files, so long as all of the files can be opened using non-ZIP64 constructs.


pkgw:
@mvdnes I have a personal project that relies on this patch; would it be possible to merge this if the change looks good, and make a new release? Thanks.

mvdnes:
Sorry for the delay. It is a great addition, many thanks! 👍

PR #48, opened 2017-10-28
Modify examples/extract.rs, reduce code.

biluohc:


mvdnes:
Although I see some improvements, I am not really happy with the new output format. Could you maybe change those?
Something like `"File {} comment: {}"` and `"File {} extracted to \"{}\" ({} bytes)"` perhaps?

biluohc:
Ok, I will modify the output format to your love.
But you use C's style to handle the Result, Maybe it is not beautiful? 
As follows:
```rust
fn main() {
    std::process::exit(real_main());
}

fn real_main() -> i32
{ 
many unwrap()
}
```

biluohc:
I use `i` as the first argument, because of the output will very long and repeated if use `file.name()`.

```rust
          println!("File {} comment: {}", i, comment);
          println!("Dir {} extracted to \"{}\"", i, outpath.as_path().display());
          println!("File {} extracted to \"{}\" ({} bytes)", i, outpath.as_path().display(), file.size());
```

biluohc:
Use `File` to replace `Dir` for aligning.
Comit ID: 09f728e
```rust
          println!("File {} comment: {}", i, comment);
          println!("File {} extracted to \"{}\"", i, outpath.as_path().display());
          println!("File {} extracted to \"{}\" ({} bytes)", i, outpath.as_path().display(), file.size());
```

mvdnes:
Sorry for the late reaction. Thanks for the change! 👍

PR #52, opened 2017-12-03
Handle invalid zip header

anvie:


mvdnes:
This should be caught indeed. Thanks! 👍

PR #54, opened 2018-01-07
Made deflate compression into separate feature

jhwgh1968:
Though likely not intended by the format's designers, my application is using ZIP files without any compression at all (the files are themselves compressed) and is targeting a platform without a C compiler (at least not one Rust can find at the moment). The most straightforward way to this goal was to disable both bzip2 and deflate compression. The latter required creating a feature gate, hence this PR.

I started this on an older crates.io version of the library, and I now see that perhaps the miniz_oxide dependency change has obsoleted this problem. If you see no other reason for this separate feature gate, feel free to just close this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ Library to support the reading and writing of zip files.
 """
 
 [dependencies]
-flate2 = { version = "1.0", default-features = false, features = ["rust_backend"]}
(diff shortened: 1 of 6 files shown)


mvdnes:
I like the idea. Could you maybe rename the feature to `deflate`? That way we would not directly be tied to the specific `flate2` crate.

mvdnes:
Thanks! 👍

PR #59, opened 2018-01-27
Improvements to zip dir example

gilescope:
* Added reference in readme to examples. (I failed to spot the examples dir)
   * Factored out the zip dir example so people could use it also to write to a memory buffer rather than a file.
   * Clears and reuses the same vec buffer rather than creating a new buffer each time.
   * Errors if source dir is not a dir rather than returning Ok.

(Ideally I'd have zip dir functionality in the crate but I can see the need to minimise dependencies.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,12 @@ Without the default features:
 [dependencies]
 zip = { version = "0.2", default-features = false }
 ```
+
+Examples
+--------
+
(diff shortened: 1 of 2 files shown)


mvdnes:
Looks good to me. Thank you for your contribution! 👍

PR #70, opened 2018-05-06
Reexport flate2's rust_backend and zlib features

baskerville:


mvdnes:
Why is this needed? Is it not sufficient to run with `--features flate2/zlib`? Or are there other cases where such a flag is not possible?

baskerville:
The flag works for `zip-rs` but not for a crate that depends on `zip-rs`.

In other words, a crate cannot set the features of a dependency of a dependency (unless the dependency re-exports its dependency's features).

mvdnes:
I have choosen to merge #73 instead, but thank you for the contribution!

PR #73, opened 2018-05-19
Control backend choice of flate2

alexbool:
This PR permits users to choose flate2's deflate backend, effectivly copying its features to zip-rs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ notifications:
 script:
     - cargo test
     - cargo test --no-default-features
+    - cargo test --no-default-features --features "deflate-zlib"
+    - cargo test --no-default-features --features "deflate-miniz"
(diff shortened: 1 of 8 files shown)


alexbool:
Whoops, I see this is basically a dupe of #70, but with some comments in README.md, and better CI.

mvdnes:
As I said with #70:
Why is this needed? Is it not sufficient to run with --features flate2/zlib? Or are there other cases where such a flag is not possible?

I do like the fact that your explanation makes it clearer for users what to do.

alexbool:
Cargo features are sort of public API of a crate, and currently users of a crate cannot set features of dependent crates. So crate authors are forced to re-export features if they want them to be configurable.

mvdnes:
Thank you for the contribution! It has been merged in 0.4.0.

Issue #60, opened 2018-01-30
Make an additional optional feature to use `deflate` and `inflate` instead of `flate2`


application-developer-DA:
I think it makes sense to provide some additional feature-guarded backends / "deflate engines" to the end user, currently `flate2` is used, but the problem with it is that it relies on some C libraries and even the `rust_backend` pulls too much C-related things and requires a refactoring. When I tried to port `zip-rs` to `WebAssembly` (Rust now has a target which allows you to compile the library as WebAssembly module) I struggled with `flate2` and its dependencies (notably it's `rust_backend` dependency) and had to replace the usage of `flate2` by the combination of `deflate` and `inflate` (that was the easiest way to go), both provide pure Rust libraries for compression and decompression. The API of `inflate` is very similar to the one provided by `flate2`, so the changes were trivial, `deflate` has however a different API and has to be adapted a bit to be able to work with `zip-rs`.

IMHO it also makes sense to enable/disable compression and decompression seperately (some people may only either compression or decompression, which is especially important for things like `WebAssembly` where you don't want to blow up the size of the compiled module).

xmclark:
@application-developer-DA did you make a fork with the refactor for `deflate` and `inflate`? I also am interested in using `zip-rs` in WebAssembly, but have not had much luck.

application-developer-DA:
@xmclark yes, I did some changes several months ago. I did not make a pull request back then, since the changes I did are not streamlined yet and it was kind of an "experimental" version. Maybe I find time some day and update the fork: https://github.com/snapview/zip-rs/tree/wasm-experiments

The fork works fine with WebAssembly. I was able to compile and use it several months ago and the compiled WASM module is still used within the project for which I did this "small experiment", I have not had any issues with it so far.

xmclark:
@application-developer-DA that is awesome! I am pretty interested getting zip on wasm, so I might follow up on your PR.

I don't quite understand how cargo features work, so I've been digging into that.

mvdnes:
With the introduction of libflate in 0.5.0, I think this issue is resolved. Feel free to reopen if this is not the case.

PR #79, opened 2018-09-25
add libflate

xmclark:
This PR addresses #60. It adds new target that uses a pure rust implementation of deflate: [libflate](https://github.com/sile/libflate). 

I think this works. I tested this by running the tests and a few of the examples compiled for the new libflate target. I also tested it with default features to make sure I didn't break anything. I did all this on Windows.

To build the new target:
```
> cargo build --no-default-features --features deflate
```

I am open to nit-pick and style changes.


xmclark:
@mvdnes can this be reviewed?

My main use case is targeting web assembly. I have noticed different strategies for conditionally compiling target specific dependencies. This repository uses the named features so I continued doing that with `sile-libflate` feature.

mvdnes:
Thank you for your contribution.
The first thing that springs to my attention is that this confuses the compression algorithm (8 = deflate) with the implementation of the encoder/decoder (flate2 / sile-libflate). I do not think this change warrants a new `CompressionMethod`.

I might be tempted to replace the old flate2 with this library, as this would ease things on multiple platforms.

xmclark:
I made it another `CompressionMethod` because I think some developers would prefer the `flate2` backend. I did not consider the confusion and I acknowledge it now.

[`flate2` is working on support for wasm](https://github.com/alexcrichton/flate2-rs/issues/161). I have no idea when that will be completed.

Adding `libflate` makes sense for me because I want to use `zip-rs` in my wasm projects. It is also a dependency in [~50 other registered crates on crates.io](https://crates.io/crates/zip/reverse_dependencies). 

I can't speak to the performance differences. So far, the `libflate` `CompressionMethod` has been meeting my expectations in my personal projects.

mvdnes:
The feature name libflate is fine, but I would rather have the enum values still be called `Deflated` (named after the algorithm) rather than `Libflate` (named after the implementation).

Also I would like the default features to include both compression methods.

src/write.rs
```diff
@@ -83,21 +87,31 @@ pub struct FileOptions {
...
             permissions: None,
         }
     }
 
-    #[cfg(not(feature = "flate2"))]
+    #[cfg(all(not(feature = "libflate"), feature = "bzip2"))]
     /// Construct a new FileOptions object
     pub fn default() -> FileOptions {
         FileOptions {
-            compression_method: CompressionMethod::Stored,
```
xmclark: @mvdnes Question about the code here:

The conditional branch for `#[cfg(feature = "bzip2")]` was not covered here. The `Stored` compression is used for bzip2 feature. Does it not matter? 

I added the edge case in my refactor. But this is currently blocking me from compiling both `deflate` and `bzip2`.

mvdnes: Yes, that is right. Altough supported here, bzip2 is not that common for zip files. That is why it must be explicitly set in `FileOption`, even when it is the only method available.

xmclark: Ok. As it is now, there is a default method for each of the cases. In the case where we build with both targets, I add an extra helper method for Bzip2. 

Would you prefer I change it back so that default case for Bzip2 is `CompressionMethod::Storage`? And there is no extra helper method when we compile both `deflate` and `bzip2` features?

mvdnes: Yes, I prefer `Stored` to be the default even if bzip2 is available.

xmclark: @mvdnes I've addressed the feedback and cleaned up the diff. I think this is ready for a little more more review!

.travis.yml
```diff
@@ -16,9 +16,9 @@ script:
       [ "$TRAVIS_RUST_VERSION" = "1.20.0" ] || cargo test
     - cargo test --no-default-features
     - |
-      [ "$TRAVIS_RUST_VERSION" = "1.20.0" ] || cargo test --no-default-features --features "deflate-zlib"
-    - cargo test --no-default-features --features "deflate-miniz"
-    - cargo doc --no-deps --no-default-features --features "deflate-miniz,bzip2"
+      [ "$TRAVIS_RUST_VERSION" = "1.20.0" ] || cargo test --no-default-features --features "deflate"
```
mvdnes: Can you check if it also works without the `[ "$TRAVIS_RUST_VERSION" = "1.20.0" ] || ` part? That was added because the previous implementation did not work on the 1.20.0 version anymore.

mvdnes: The part `[ "$TRAVIS_RUST_VERSION" = "1.20.0" ] ||` is a check for the rust version. If the version matches, then it returns true without even running the test.
Since libflate seems to run fine on version 1.20.0, we do not need to skip the tests anymore. I think the travis script can become this:

```yaml
    - cargo test
    - cargo test --no-default-features
    - cargo doc --no-deps
    - rustdoc --test README.md -L target/debug
```

xmclark: great! I removed the additional checks.

appveyor.yml
```diff
@@ -42,4 +42,5 @@ build: false
 test_script:
   - cargo test
   - cargo test --no-default-features
-  - cargo test --no-default-features --features "deflate-miniz"
+  - cargo test --no-default-features --features "deflate"
+  - cargo test --no-default-features --features "bzip2"
```
xmclark: @mvdnes you made comments on the travis build. Should appveyor build each feature individually?

xmclark: I've removed the extra builds in appveyor.

mvdnes:
Thanks for all the work!
I am going to review some more of the public interface before publishing the 0.5.0 version, but hope to do that soon!

alexbool:
> Seeing as libflate is more 'self-contained' I think it would be beneficial to the project to use just that. This would also be a good time to remove all the (possibly confusing) features of different flate flavours.

@mvdnes while having a pure-rust implementation for deflate as a default is a step forward a clear no-brainer, removing optional `flate2` backend, in my opinion, is suboptimal. Before that one could use `flate2`'s capability to use and link to system `zlib`, now this is gone. Could we bring `flate2` as an optional backend back?

alexbool:
@mvdnes

xmclark:
@mvdnes would you be interested in opening an issue for the plugin system? I am interested in following the progress and maybe helping if I can.

PR #80, opened 2018-10-02
Replace `try!` macros with `?` operator calls

bfrascher:
The `?` operator exists since Rust version 1.13.0 and has since become the
standard and recommended variant over the `try!` macro (see
https://doc.rust-lang.org/std/macro.try.html where it is explicitly mentioned to
use the `?` operator instead of the `try!` macro).

I think it is especially useful to replace the `try!` usages throughout the
examples (since new users might not be familiar with the `try!` macro at all).


mvdnes:
Thank you for the contribution! :+1:

PR #81, opened 2018-10-04
Add `chrono` as an optional crate to replace `time` and `msdos_time`

bfrascher:
The `time` crate is no longer actively maintained.  It lists `chrono` as an alternative, and looking at the 'Recent' statistics on crates.io `chrono`, seems to be about as popular as `time`.  For all the `chrono` users out there it would be nice, not to have to convert between `time`'s structs and `chrono`'s structs whenever working with zip files.

The disadvantage of the PR is that it makes the code base quite a bit more complex with a lot of `#[cfg]` attributes.  I am especially unhappy with the added nested code blocks in `read.rs` required to guard multiple code lines with a single `#[cfg]` attribute, but I could not find a better solution.  The `cfg!` macro does not help, because both code paths must be valid with any set of feature flags.  This does not work out, because the one branch needs `chrono` types (which are only available with the 'chrono' feature flag) and the other ones requires `time` types (which are only available if the 'chrono' feature flag has __not__ been set).

Finally I pulled in the `#[cfg]` attribute for `write::FileOptions::default`, so that we don't need to define the whole function twice.  Technically this does not directly have anything to do with the `chrono` addition, but made my life a bit easier since I didn't have to add the guarded definitions of `last_modified_time` to both versions of the method.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,11 @@ Library to support the reading and writing of zip files.
 """
 
 [dependencies]
+chrono = { version = "0.4", optional = true }
 flate2 = { version = "1.0", default-features = false, optional = true }
-time = "0.1"
+time = { version = "0.1", optional = true }
 podio = "0.1"
-msdos_time = "0.1"
+msdos_time = { version = "0.1", optional = true }
 bzip2 = { version = "0.3", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


bfrascher:
I seem to have forgotten about the travis and appveyor test configurations.  When using `--no-default-features` one now has to also add either 'time' and 'msdos_time' or 'chrono' to the features list.  I don't know of a better to do this.  One would need a way to enable some features if another feature is disabled, but I don't think cargo supports that (yet).

Should I simply update the test definitions (and include one for chrono while I'm at it)?

mvdnes:
The tests are fixed, that should not be an issue anymore.

I am unsure what to do with this. I definitely do not like the `cfg`'s, but I find it hard to choose between time and chrono. If I would build zip-rs at this moment, I think chrono would be the preferred choice, but at this stage I do not see any great benefit in changing it.

mvdnes:
I have added a custom `DateTime` structure in version 0.5, with optional support for conversion to `time::Tm`. If you feel the need, you may add a similar feature for chrono in a new PR.

Issue #82, opened 2018-10-08
Additional tests


camchenry:
I was thinking about implementing additional tests and I wanted to get some feedback before implementing anything.

First one I was thinking about was a circular zip test.

1. Zip some arbitrary data with zip-rs
2. Unzip that data with zip-rs
3. Assert that start data == end data

This seems like something that should **always** be true, no matter what. It also seems like a usability issue if zip-rs was somehow able to create a zip file that it couldn't read. This would also provide a good test for fuzzing in the near future.

Additional test ideas are welcome. Thanks.

mvdnes:
Such a test would be a great idea. Feel free to implement it.

camchenry:
Ok, I will try to implement it when I have some time.

PR #83, opened 2018-10-13
Add end-to-end test

camchenry:
This implements the test described in #82. I'm still new to Rust, so please give me feedback on how it looks and functions. I based it on the examples, just combining the write and extract samples.


camchenry:
Looks like the build is failing because the buffer is overflowing, since it's now tied to the length of the data being written. Any thoughts on what I should change? Increasing the buffer size causes the zip data to be unreadable.

camchenry:
Okay, I have made those changes. Just waiting on the tests to pass now.

mvdnes:
Thanks! :+1:

PR #86, opened 2018-11-06
File info perf improvements

srijs:
Hi!

I'm using the `zip` crate to load rather big (50+MB, 10K+ files) zip archives, identify a handful of files that are of interest to me, and extract those.

While benchmarking my application, I noticed that a large amount of time (~40%) is spent in the `ZipArchive::new` method.

In order to improve the performance of this operation, I have implemented changes to defer two relatively costly operations:

- Deferring the parsing of the local file header until `ZipArchive::by_index` is called helps with the efficiency of `BufReader`, since it eliminates 2 seek operations per file entry which used to invalidate the read buffer and force a syscall
- Deferring the conversion of the MSDOS datetime to `time::Tz` until `ZipFile::last_modified` is called, since `Tm::to_timespec` invokes a syscall (at least on my OSX machine) every time it is called (once per file entry)

Implementing these two changes reduced the time taken for the `file_info` example from 380-400ms down to 210-230ms (w/o using `BufReader` the initial runtime was ~650ms).

The two optimisations are largely independent, so let me know if you'd like me to split this into multiple PRs!

For reference, an `unzip -l` on the same file takes about 140ms, so I'm planning to investigate further and see what other improvements can be made.

Let me know what you think!

### DIFF
--- a/examples/file_info.rs
+++ b/examples/file_info.rs
@@ -1,6 +1,7 @@
 extern crate zip;
 
 use std::fs;
+use std::io::BufReader;
 
 fn main() {
     std::process::exit(real_main());
@@ -14,8 +15,9 @@ fn real_main() -> i32 {
     }
     let fname = std::path::Path::new(&*args[1]);
     let file = fs::File::open(&fname).unwrap();
+    let reader = BufReader::new(file);
 
-    let mut archive = zip::ZipArchive::new(file).unwrap();
+    let mut archive = zip::ZipArchive::new(reader).unwrap();
 
     for i in 0..archive.len() {
         let file = archive.by_index(i).unwrap();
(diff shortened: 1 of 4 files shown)


mvdnes:
I like your ideas!

The only problem I see is that the `data_start` is now calculated at a later time. I think one solution is to calculate it on-demand, when calling the `data_start()` method.

In a later version of this library, I want to change the datetime objects as it was indeed a bit slow for how simple these fields are. For now though, and for compatibility, your changes are very welcome.

srijs:
Hmm, given that `data_start` is calculated in `ZipArchive::by_index` _before_ constructing the `ZipFile`, at the point where someone calls `ZipFile::data_start()` it should already be returned correctly. Or have I misunderstood your concern?

mvdnes:
Yes you are right. No changes needed then.
Thanks for the contribution! :+1:

PR #87, opened 2018-11-07
Improve CRC32 throughput

srijs:
After tackling #86, the CRC32 reader started to become a bottleneck when extracting files from the archive.

To improve the throughput of the CRC32 function, I largely followed [Stefan Brumme's write-up on fast CRC32 algorithms](https://create.stephan-brumme.com/crc32/) which lead me to the current implementation which can process 64 bytes of input in a single iteration at the cost of ~16KB of static memory to host the constant tables.

There is also a potential implementation using `pclmulqdq` instructions on newer x86_64 processors, but my thinking so far was to improve the performance for all architectures first, before considering architecture-specific code paths.

To benchmark the implementation I've been using a zipfile with a PDF copy of ["A History of Haskell:
Being Lazy With Class"](http://haskell.cs.yale.edu/wp-content/uploads/2011/02/history.pdf) which is roughly 1MB in size. In this benchmark, my changes result in a speedup of ~1.5X (from 12.6ms down to 8.2ms per iteration).

After these changes, deflate has now moved to be the main bottleneck with about 90% of cpu time going towards the decompression functions.

I briefly considered whether this additional logic should go into a separate crate, but I thought I'd just open the PR as-is and we can discuss here.

If you are concerned about the increase in static memory (embedded, etc), it could also be an option to feature-flag this.

Looking forward to your feedback!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ podio = "0.1"
 msdos_time = "0.1"
 bzip2 = { version = "0.3", optional = true }
 libflate = { version = "0.1.16", optional = true }
+crc32fast = "1.0"
 
 [dev-dependencies]
 bencher = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,14 @@ bzip2 = { version = "0.3", optional = true }
 libflate = { version = "0.1.16", optional = true }
 
 [dev-dependencies]
+bencher = "0.1"
+rand = "0.4"
 walkdir = "1.0"
 
 [features]
 deflate = ["libflate"]
 default = ["bzip2", "deflate"]
+
+[[bench]]
+name = "read_entry"
+harness = false
--- a/benches/read_entry.rs
+++ b/benches/read_entry.rs
@@ -1,0 +1,46 @@
+#[macro_use]
+extern crate bencher;
+extern crate rand;
+extern crate zip;
+
+use std::io::{Cursor, Read, Write};
+
+use bencher::Bencher;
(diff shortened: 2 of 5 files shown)


mvdnes:
I think this would be useful as a seperate crate, but I do not mind including it directly.

Could you include the zip that is being benchmarked? And please make sure that it is created without compression (using `zip -0`), so that the checksumming can be compared on itself.

srijs:
I've modified the benchmark to generate a random 1MB zip during runtime so that it doesn't need to depend on any fixtures. I can check in the fixture instead if you really want me to, I was just keen on finding a way to not commit a large binary file.

With compression turned off, this is the new benchmark result:

```
 name        old ns/iter           new ns/iter          diff ns/iter   diff %  speedup
 read_entry  3,787,231 (276 MB/s)  658,477 (1592 MB/s)    -3,128,754  -82.61%   x 5.75
```

srijs:
No problem, instead of checking in the CRC32 code directly, I've changed this now to consume my [crc32fast](https://crates.io/crates/crc32fast) crate.

In the meantime I've also managed to implement the `pclmulqdq` optimizations, so I'm happy to report that with the new changes this results in a cool **30X** speed-up in the `read_entry` benchmark:

```
 name        old ns/iter           new ns/iter          diff ns/iter   diff %  speedup
 read_entry  3,743,605 (280 MB/s)  120,611 (8693 MB/s)    -3,622,994  -96.78%  x 31.04
```

mvdnes:
Thanks for this! :+1:

PR #89, opened 2018-11-23
Read files inside ZipArchive lazily

Arnavion:
Before this change, the entire central directory would be parsed to extract
the details of all files in the archive up-front.

With this change, the central directory is parsed on-demand - only enough
headers are parsed till the caller finds the file it wanted.

This speeds up callers who only wish to pick out specific files from
the archive, especially for archives with many thousands of files or
archives hosted on slow storage media. In the case where the caller *did* want
every file in the archive, this performs no worse than the original code.


srijs:
Nice! Just tried out this patch on a project I'm working on, which is essentially extracting a handful of files from a very large archive, and it reduced total runtime by ~30% from 180-200ms down to 120-130ms.

PR #91, opened 2018-12-05
Zero copy central directory parsing

srijs:
This changeset is based on #89, and aims to improve performance further by minimising allocations and memcpy's when parsing central directory records.

In a nutshell, it reads the full central directory into a buffer, and then relies on the changes in #89 to incrementally parse the bytes. While parsing, filenames and comments are made to share the underlying data of the main buffer, avoiding any additional allocations and copies.

Overall this improves performance for the `file_info` example by more than 25% (for a ~60MB, 20K entries zip file from ~470ms down to ~340ms).

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,0 +1,128 @@
+use std::borrow::{Borrow, Cow};
+use std::io::Read;
+use std::ops::Deref;
+use std::hash::{Hash, Hasher};
+
+use bytes::{Buf, Bytes};
+use string;
+
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct StrBuf(string::String<Bytes>);
+
+impl StrBuf {
+    pub(crate) fn from_str(s: &str) -> Self {
+        StrBuf(string::String::from_str(s))
+    }
+
+    pub(crate) fn from_utf8(bytes: ByteBuf) -> Result<Self, ::std::str::Utf8Error> {
+        Ok(StrBuf(string::TryFrom::try_from(bytes.0)?))
+    }
+
+    pub(crate) fn from_utf8_lossy(bytes: ByteBuf) -> Self {
+        match String::from_utf8_lossy(bytes.as_ref()) {
+            Cow::Owned(s) => s.into(),
+            Cow::Borrowed(s) => {
+                // SAFETY: We know that `bytes` only contains valid utf-8,
+                //         since the `from_utf8_lossy` operation returned the
+                //         input verbatim.
+                debug_assert_eq!(s.len(), bytes.len());
+                StrBuf(unsafe { string::String::from_utf8_unchecked(bytes.clone().0) })
+            }
+        }
+    }
+}
+
+impl From<String> for StrBuf {
+    fn from(s: String) -> Self {
+        let bytes = s.into_bytes().into();
+        // SAFETY: We know that `bytes` only contains valid utf-8,
+        //         since the underlying data comes from the input string.
+        StrBuf(unsafe { string::String::from_utf8_unchecked(bytes) })
+    }
+}
+
+impl Hash for StrBuf {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        // Because of the impl Borrow<str> for StrBuf, we need to make sure that the Hash
+        // implementations behave identically between str and StrBuf.
+        //
+        // Quoting the documentation for the Borrow trait:
+        //
+        // > Further, when providing implementations for additional traits, it needs to be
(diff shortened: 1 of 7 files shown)


srijs:
(the build failures appear to be due to `libflate` failing to compile with Rust 1.20)

src/read.rs
```diff
@@ -301,45 +299,45 @@ impl<R: Read+io::Seek> ZipArchive<R>
...
+    let signature = bytes.read_u32::<LittleEndian>()?;
     if signature != spec::CENTRAL_DIRECTORY_HEADER_SIGNATURE
     {
         return Err(ZipError::InvalidArchive("Invalid Central Directory header"))
     }
 
-    let version_made_by = reader.read_u16::<LittleEndian>()?;
-    let _version_to_extract = reader.read_u16::<LittleEndian>()?;
-    let flags = reader.read_u16::<LittleEndian>()?;
+    let version_made_by = bytes.get_u16_le();
```
Arnavion: What if the central directory header lied and the size it reports is too small? Then the central directory buffer might end in the middle of a record, and one of these `get_*` will panic. The original `read_*` would have read garbage (bits of the end-of-file record) or return `Err(UnexpectedEof)`

Since you already gave it a `Read` impl and still use the original `read_u32` on line 305, maybe keep the `read_*` here as well?

src/buffer.rs
```diff
@@ -0,0 +1,113 @@
...
+                debug_assert_eq!(s.len(), bytes.len());
+                StrBuf(unsafe { string::String::from_utf8_unchecked(bytes.clone().0) })
+            }
+        }
+    }
+}
+
+impl From<String> for StrBuf {
+    fn from(s: String) -> Self {
+        let ByteBuf(bytes) = ByteBuf::from(s.into_bytes());
```
Arnavion: Why not create the `bytes::Bytes` directly from `s.into_bytes()` with `.into()`, instead of round-tripping it through `ByteBuf::from` ? (Though it would get optimized the same anyway.)

src/buffer.rs
```diff
@@ -0,0 +1,113 @@
...
+        // SAFETY: We know that `bytes` only contains valid utf-8,
+        //         since the underlying data comes from the input string.
+        StrBuf(unsafe { string::String::from_utf8_unchecked(bytes) })
+    }
+}
+
+impl Hash for StrBuf {
```
Arnavion: Why not `#[derive]` it? Is there a benefit to matching the `str`'s hash rather than the `Bytes`'s hash (which `String<Bytes>`'s `Hash` forwards to)?

srijs: Because of the `impl Borrow<str> for StrBuf`, I need to make sure that the hash implementations behave identically between `str` and `StrBuf`.

Quoting the documentation for the [`Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) trait:

> Further, when providing implementations for additional traits, it needs to be considered whether they should behave identical to those of the underlying type as a consequence of acting as a representation of that underlying type. Generic code typically uses `Borrow<T>` when it relies on the identical behavior of these additional trait implementations. These traits will likely appear as additional trait bounds.

Without this, it would be impossible to look up an entry from the `names_map` by `&str`, since the `str` and `StrBuf` would evaluate to different hashes, even if they represent the same sequence of chars.

src/buffer.rs
srijs: Good catch, this must be a leftover of a refactor I did. I'll get rid of it!

src/read.rs
srijs: That would be possible, but then we're still left with the calls to `split_to` for which `Read` doesn't have an  equivalent. Maybe the way to do it would be to check the length at the start of the method, and return a `Err(UnexpectedEof)` if it is too short.

Arnavion: >That would be possible, but then we're still left with the calls to `split_to` for which `Read` doesn't have an equivalent.

Yes, you'd still have to check `.remaining()` manually for the three variable-length fields before `split_to()`ing them.

>Maybe the way to do it would be to check the length at the start of the method

Yes, this would be the most efficient (avoids the bound checks), though you'd still need to check the three variable-length fields separately from the constant-sized header.

srijs: Yep, cool, that's what I meant. I'll go and make these changes!

srijs:
Hi! This PR has been up for a while, and I‘ve stopped working on the project I needed this for.  

I’m happy for someone else to take the baton here if you‘re interested in improving zip parsing performance, but unfortunately I don‘t have the time or interest to carry this forward at the moment.

Plecra:
Thanks for the contribution ❤️ I'm sure it will factor into the more streaming-oriented design we're working on.

PR #94, opened 2019-01-12
Make traits for zip files

jmerdich:
It's currently very difficult to have a list or struct of ZipArchives with different underlying reader types (eg. some file-backed, some memory-backed). This introduces new traits for the structs that have generic arguments. 

Eg, old:
```rust
struct MyStructThatHasAZip<T> {
   zip : Box<ZipArchive<T>>;
}
```
vs new:
```rust
struct MyStructThatHasAZip {
   zip : Box<ZipArchiveRead>;
}
```

 **This would be a breaking API change since any method use would require importing the relevant trait instead of the archive type. This is a rust limitation/"feature"**

Possible improvements: 
 - Make ZipWriter implement ZipArchiveRead to allow reading back data

### DIFF
--- a/examples/extract.rs
+++ b/examples/extract.rs
@@ -2,6 +2,7 @@ extern crate zip;
 
 use std::io;
 use std::fs;
+use zip::ZipArchiveRead;
 
 fn main() {
     std::process::exit(real_main());
--- a/examples/extract_lorem.rs
+++ b/examples/extract_lorem.rs
@@ -1,4 +1,5 @@
 use std::io::prelude::*;
+use zip::ZipArchiveRead;
 
 extern crate zip;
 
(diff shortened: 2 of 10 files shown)


mvdnes:
I understand the idea, but is this not a bit a niche problem? For the small use-cases that do want this functionality, it could be done by creating an own trait.

Feel free to disagree, I would love to hear some more information.

Plecra:
There's also the option of boxing the reader used by `zip` if you're looking for a concise solution.
```
struct MyStructThatHasAZip {
   zip : ZipArchive<Box<dyn Read + Seek>>;
}
```

This PR is quite stale now, and won't get attention in the short term. Feel free to submit a new one if you feel there's still a need for this 😀

PR #96, opened 2019-01-27
Add ZipFile::is_dir() and ZipFile::is_file().

### LINKED ISSUE #69: Add ZipFile.is_directory()

damiencollard:
Fixes #69.

The naming matches that of std::fs::Metadata.

An entry is determined to be a directory based on the presence of
a trailing path separator, i.e. '/' or '\\'.

This patch adds a small test zip containing files and directories.
Their names match their type so as to make testing easy.


mvdnes:
Good idea, thanks! :+1:

syohex:
https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

>      4.4.17.1 The name of the file, with optional relative path.
>      The path stored MUST NOT contain a drive or
>     device letter, or a leading slash.  All slashes
>     MUST be forward slashes '/' as opposed to
>       backwards slashes '\' for compatibility with Amiga
>     and UNIX file systems etc.  If input came from standard
>     input, there is no file name field.  

Zip specification says that path separator MUST be  forward slash(`/`). So should `is_dir` method check only a forward slash ?

PR #100, opened 2019-03-06
Fix bounds check in DateTime::from_time

nareshganduri:
The bounds check in the DateTime::from_time function did not use the correct bounds for checking the year.

The new bounds are [80, 207], which accounts for the tm_year field storing the years since 1900.


mvdnes:
Nice find, thanks! 👍

PR #101, opened 2019-03-20
Add basic iterator/find support

Jake-Shadle:
ZipArchive is a bit cumbersome to use when you are consuming archives with unknown/variable contents, so this PR exposes 3 helper methods on ZipArchive to make it slightly easier to consume such content.

* `iter_files()` - Returns an iterator over the internal vector of ZipFileData, which allows people to use all of the standard iterator goodies they are used to
* `by_id()` - As with the existing by_name/by_index, returns the ZipFile if that id exists. The id can be converted from a &ZipFileData and is unique (not name based)
* `find()` - While iter_files().map() + by_id() or iter_files().position() + by_index() can be used to find a file matching criteria, they are a bit cumbersome to use due to the borrow checker since iter_files() is an immutable borrow and all of by_ methods are mutable, so `find` just wraps the most common case of "get me the file that matches this predicate, if it exists"

### DIFF
--- a/src/cp437.rs
+++ b/src/cp437.rs
@@ -170,7 +170,6 @@ fn to_char(input: u8) -> char
         0xfd => 0x00b2,
         0xfe => 0x25a0,
         0xff => 0x00a0,
-        _ => unreachable!(),
     };
     ::std::char::from_u32(output).unwrap()
 }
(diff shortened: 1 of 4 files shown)


repi:
Any updates or ideas on this to make it acceptable and merged in?

mvdnes:
I am sorry, but due to a lack of time I don't think I will be able to review anything from this library anytime soon.

steven-joruk:
Sorry for going off topic Jake, I would like to see the iter part of this MR go in - @mvdnes would you consider adding other maintainers?

PR #102, opened 2019-03-30
Use add_directory in write_dir example

jonpas:
Fix compatibility with some unzip tools. Comments should explain everything.

Example itself is untested, but implementation is confirmed working - I basically copied what I had to do in https://github.com/synixebrett/HEMTT/pull/9 to form a properly formatted archive.


mvdnes:
Good idea, thanks! :+1:

PR #104, opened 2019-04-10
add support for concurrently read on ZipFile level

### LINKED ISSUE #14: concurrent reads
Hello Maintainer,

Firstly, thank you for creating and maintaining this library!

Secondly, are you interested in adapting this library to support concurrent reads?

For my application, I need to read entries (at random) from a zip file over an extended period of time. The reads might be concurrent, and might be interleaved (read a few bytes from zip entry A, then a few bytes from zip entry B, then A again, etc).

The current API shape makes this use-case impossible without creating multiple `ZipArchive`s.

I've [hacked the code to accomplish my goals](https://github.com/booktrack/zip-rs/compare/master...booktrack:concurrent-read), and thought I'd share it here in case you or any library users are interested in doing something similar.

I'd rather not maintain a separate fork of this library just to suit my use-case if it can be avoided, so if there is interest in incorporating any kind of support for concurrent reads, I'm happy to help out.

If there is no interest, please don't hesitate to close this issue.

clouds56:
* note: the PR fixes #14 and contains API break changes
* add trait `ZipFileEntry` that `ZipFile` should implement
* add/modify struct `ZipFile`, `ZipStreamFile` and `ZipFileOwned` for seekable reader, stream reader and reader that cloned.
* add `ZipFile::try_clone` which returns a new `ZipFileOwned`

### DIFF
--- a/examples/extract.rs
+++ b/examples/extract.rs
@@ -2,6 +2,7 @@ extern crate zip;
 
 use std::io;
 use std::fs;
+use zip::read::ZipFileEntry;
 
 fn main() {
     std::process::exit(real_main());
--- a/examples/file_info.rs
+++ b/examples/file_info.rs
@@ -2,6 +2,7 @@ extern crate zip;
 
 use std::fs;
 use std::io::BufReader;
+use zip::read::ZipFileEntry;
 
 fn main() {
     std::process::exit(real_main());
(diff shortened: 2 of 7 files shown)


BenjaminRi:
@mvdnes Can we please move forward on some of these pull requests? The more there are, the harder they become to merge. This is a feature that is very useful.

Plecra:
There are a few things I'd want to touch up in this PR, but it's gone stale. Feel free to open a new issue if anyone would like to revisit this, but I'd like to talk about the API before implementation work is done 😃 .

PR #107, opened 2019-06-14
Switch to flate2

### LINKED ISSUE #88: Comparatively poor performance (both speed & size)
Compared to the `zip` command-line tool, or compared to the standard OpenJDK implementation (which is equivalent), `zip-rs` has poor performance in both senses: it is slow (~3x slower than the standard implementations and equivalent compression), and it has only approximately the worst deflate compression available with zip.

Since compression can be a rate-limiting operation (in my case it is), this is a rather significant drawback.  I realize that a significant part of this is due to deficiencies in `libflate`, but it still is a substantial negative for this crate.  If a native implementation that performs acceptably is too tricky to create, a feature wrapping `libzip` would render the crate usable.

If you find some file of a dozen MB or so and name it `speed.log`, the following will test the performance:

```rust
#[derive(Debug, Clone, Copy)]
pub struct CompressionResult {
    pub dt: f64,
    pub MBps: f64,
    pub shrink: f64
}
impl CompressionResult {
    pub fn new(dt: f64, MBps: f64, shrink: f64) -> CompressionResult {
        CompressionResult { dt, MBps, shrink }
    } 
    pub fn from(dt: f64, old_size: u64, new_size: u64) -> CompressionResult {
        let MBps = ((old_size as f64)/(1024.0*1024.0))/dt;
        let shrink = (new_size as f64)/(old_size as f64);
        CompressionResult {
            dt: (dt*1000.).round()/1000.,
            MBps: (MBps*100.).round()/100.,
            shrink: (shrink*1000.).round()/1000.
        }
    }
}

pub fn compress_the_target(cs: &str, ct: &str) -> CompressionResult {
    use std::*;
    use std::io::Write;
    let source_p = path::Path::new(cs);
    let target_p = path::Path::new(ct);
    if target_p.exists() { fs::remove_file(target_p); }
    let data = fs::read(source_p).unwrap();
    let t0 = time::Instant::now();
    {
        let target = fs::File::create(target_p).unwrap();
        let buffer = io::BufWriter::with_capacity(65536, target);
        let mut zw = zip::ZipWriter::new(buffer);
        zw.start_file(
            source_p.file_name().unwrap().to_str().unwrap(),
            zip::write::FileOptions::default().compression_method(
                zip::CompressionMethod::Deflated
            )
        );
        zw.write(data.as_ref());
        zw.finish();
    }
    let elapsed = {
        let d = t0.elapsed();
        (d.as_secs() as f64) + (d.subsec_nanos() as f64)/1e9
    };
    let old_size = fs::metadata(source_p).unwrap().len();
    let new_size = fs::metadata(target_p).unwrap().len();
    
    CompressionResult::from(elapsed, old_size, new_size)
}

fn main() {
    let compression_source = "speed.log";
    let compression_target = "speed-rust.zip";
    println!("{:?}", compress_the_target(compression_source, compression_target));
}
```

You can get the same report for the command-line version by calling it as an external process, e.g. here in Python:

```python
import time
import subprocess
import os

cmdline_compression_source = 'speed.log'
cmdline_compression_target = 'speed-cmdline.zip'

def compress_the_target():
    if os.path.isfile(cmdline_compression_target):
        os.remove(cmdline_compression_target)
    t0 = time.time()
    subprocess.run(['zip', '-7', cmdline_compression_target, cmdline_compression_source])
    elapsed = time.time() - t0
    log_size = os.stat(cmdline_compression_source).st_size
    zip_size = os.stat(cmdline_compression_target).st_size
    rate = (log_size/(1024*1024.0))/elapsed
    factor = zip_size/log_size
    os.remove(cmdline_compression_target)
    return (round(elapsed*1000)/1000, round(rate*100)/100, round(factor*1000)/1000)

compress_the_target()
```

jonpas:
Closes #88.

This pull request switches back to [`flate2`](https://github.com/alexcrichton/flate2-rs) implementation of (DE)FLATE, from [`libflate`](https://github.com/sile/libflate). `flate2` is ahead of `libflate`, ref. https://github.com/mvdnes/zip-rs/issues/88#issuecomment-491185791.

~~Also adds default value in `cp437.rs` since newer `rustc` versions error on missing `_` resolution.~~ Was using older `rustc` by mistake.

## Benchmarks

All benchmarks were ran on `release` build (`cargo build --release`) of a program using `zip-rs` - [HEMTT](https://github.com/synixebrett/HEMTT/pull/134). Code is using `BufReader`, `BufWriter` and `copy` (instead of buffer) - code below.

On Arch Linux 5.1.8. Only zipping in the program was timed (using `std::time`).

_**ACE3** Folder Size is is different due to different build process on Windows - irrelevant for this._

### Linux

Project | Folder Size  | `libflate` Speed | `flate2` Speed | `libflate` Zip Size | `flate2` Zip Size
-------- | ------- | ----- | ----- | ------- | -------
**TACU** | 12.2 MB | 1.5 s | 0.5 s | 11.2 MB | 11.2 MB
**TACR** | 59.2 MB | 7.5 s | 2.5 s | 50.6 MB | 50.5 MB
**ACE3** | 205 MB  | 15 s  | 7.2 s | 143 MB  | 139 MB

### Windows

Project | Folder Size | `libflate` Speed | `flate2` Speed | `libflate` Zip Size | `flate2` Zip Size
-------- | ------- | ----- | ----- | ------- | -------
**TACU** | 13 MB   | 2.5 s | 0.5 s | 11.2 MB | 11.1 MB
**TACR** | 59.2 MB | 12 s  | 3 s   | 50.6 MB | 50.4 MB
**ACE3** | 174 MB  | 30 s  | 8.5 s | 137 MB  | 135 MB

## Benchmark Code

Excerpt from the program, where it is timed.

```rust
let now = std::time::Instant::now();

let zippath = Path::new(&zipsubpath);
let file = BufWriter::new(File::create(&zippath).unwrap());

let dir = walkdir::WalkDir::new(&release_dir);
let mut zip = zip::ZipWriter::new(file);
let options = zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Deflated);

// Zip all files and folders in all subdirectories
for entry in dir.into_iter().filter_map(|e| e.ok()) {
    let path = entry.path();
    let name = path.strip_prefix(Path::new(&release_dir)).unwrap();

    // Write file or directory explicitly
    // Some unzip tools unzip files with directory paths correctly, some do not!
    if path.is_file() {
        zip.start_file_from_path(name, options)?;

        let mut f = BufReader::new(File::open(path)?);

        // Copy directly, without any buffer, as we have no use for the intermediate data
        copy(&mut f, &mut zip)?;
    } else if name.as_os_str().len() != 0 {
        // Only if not root! Avoids path spec / warning
        // and mapname conversion failed error on unzip
        zip.add_directory_from_path(name, options)?;
    }
}
zip.finish()?;

println!("{}.{} seconds", now.elapsed().as_secs(), now.elapsed().subsec_nanos());
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ rand = "0.4"
 walkdir = "1.0"
 
 [features]
-deflate = ["flate2", "flate2/default"]
-deflate-zlib = ["flate2", "flate2/zlib"]
-deflate-rust = ["flate2", "flate2/rust_backend"]
+deflate = ["flate2/default"]
+deflate-zlib = ["flate2/zlib"]
+deflate-rust = ["flate2/rust_backend"]
 default = ["bzip2", "deflate", "time"]
 
 [[bench]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,9 @@ rand = "0.4"
 walkdir = "1.0"
 
 [features]
-deflate = ["flate2/default"]
+deflate = ["flate2/rust_backend"]
+deflate-miniz = ["flate2/default"]
 deflate-zlib = ["flate2/zlib"]
-deflate-rust = ["flate2/rust_backend"]
 default = ["bzip2", "deflate", "time"]
 
 [[bench]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,10 @@ Library to support the reading and writing of zip files.
 """
 
 [dependencies]
+flate2 = { version = "1.0", default-features = false, optional = true }
 time = { version = "0.1", optional = true }
 podio = "0.1"
 bzip2 = { version = "0.3", optional = true }
-libflate = { version = ">=0.1.21", optional = true }
 crc32fast = "1.0"
 
 [dev-dependencies]
@@ -24,7 +24,9 @@ rand = "0.4"
 walkdir = "1.0"
 
 [features]
-deflate = ["libflate"]
+deflate = ["flate2", "flate2/default"]
+deflate-zlib = ["flate2", "flate2/zlib"]
+deflate-rust = ["flate2", "flate2/rust_backend"]
 default = ["bzip2", "deflate", "time"]
 
 [[bench]]
--- a/examples/write_dir.rs
+++ b/examples/write_dir.rs
@@ -17,9 +17,9 @@ fn main() {
 
 const METHOD_STORED : Option<zip::CompressionMethod> = Some(zip::CompressionMethod::Stored);
 
-#[cfg(feature = "flate2")]
+#[cfg(feature = "deflate")]
 const METHOD_DEFLATED : Option<zip::CompressionMethod> = Some(zip::CompressionMethod::Deflated);
-#[cfg(not(feature = "flate2"))]
+#[cfg(not(feature = "deflate"))]
 const METHOD_DEFLATED : Option<zip::CompressionMethod> = None;
 
 #[cfg(feature = "bzip2")]
--- a/examples/write_dir.rs
+++ b/examples/write_dir.rs
@@ -17,9 +17,9 @@ fn main() {
 
 const METHOD_STORED : Option<zip::CompressionMethod> = Some(zip::CompressionMethod::Stored);
 
-#[cfg(feature = "deflate")]
+#[cfg(feature = "flate2")]
 const METHOD_DEFLATED : Option<zip::CompressionMethod> = Some(zip::CompressionMethod::Deflated);
-#[cfg(not(feature = "deflate"))]
+#[cfg(not(feature = "flate2"))]
 const METHOD_DEFLATED : Option<zip::CompressionMethod> = None;
 
 #[cfg(feature = "bzip2")]
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -9,7 +9,7 @@ pub enum CompressionMethod
     /// The file is stored (no compression)
     Stored,
     /// Deflate in pure rust
-    #[cfg(feature = "flate2")]
+    #[cfg(feature = "deflate")]
     Deflated,
     /// File is compressed using BZIP2 algorithm
     #[cfg(feature = "bzip2")]
@@ -23,7 +23,7 @@ impl CompressionMethod {
     pub fn from_u16(val: u16) -> CompressionMethod {
         match val {
             0 => CompressionMethod::Stored,
-            #[cfg(feature = "flate2")]
+            #[cfg(feature = "deflate")]
             8 => CompressionMethod::Deflated,
             #[cfg(feature = "bzip2")]
             12 => CompressionMethod::Bzip2,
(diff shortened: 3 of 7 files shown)


mvdnes:
Thank you for your contribution. However, I do not think this is the way to go as some people building for wasm were not able to use flate2.

I was working on a branch for allowing people to choose their own library as needed, but have not found the time to finish it. See the *plugin* branch in this repository.

Some further questions:
* Why do you re-add the `_` in cp437? In the current stable and nightly it works fine without it.
* If I where to merge this, the feature name should still be named `deflate' which maps to `flate2'. I do not want the external library name to leak unnecessarily.

jonpas:
flate2 should work for WASM, ref. https://github.com/alexcrichton/flate2-rs/issues/161.

> Why do you re-add the _ in cp437? In the current stable and nightly it works fine without it.

Because with `rustc 1.32.0 (9fda7c223 2019-01-16)`:
```
error[E0004]: non-exhaustive patterns: `_` not covered
  --> src/cp437.rs:42:24
   |
42 |     let output = match input
   |                        ^^^^^ pattern `_` not covered

```

> If I where to merge this, the feature name should still be named deflate' which maps to flate2'. I do not want the external library name to leak unnecessarily.

Done.

Shnatsel:
I am now making an auditing pass over libflate, and its memory safety story is not great. Here are just the bugs I've found:
https://github.com/sile/libflate/issues/16
https://github.com/sile/libflate/issues/29
https://github.com/sile/libflate/issues/31
https://github.com/sile/libflate/issues/33
I'm not a security professional and may miss bugs, so even after I'm done you should not rely on it being correct.

Cargo.toml
```diff
@@ -24,7 +24,9 @@ rand = "0.4"
 walkdir = "1.0"
 
 [features]
-deflate = ["libflate"]
+deflate = ["flate2/default"]
+deflate-zlib = ["flate2/zlib"]
+deflate-rust = ["flate2/rust_backend"]
```
Shnatsel: If you want to make this a non-breaking change, rust backend should be the default so you would not require users to have a C-to-WASM compiler to build it.

AFAIK there is no reason to not use the Rust backend for flate2 in 2019. Last I looked at it, performance was on par with or slightly better than C, and it was in good shape safety-wise.

Shnatsel: Upon closer inspection, the safety of the Rust backend for flate2 is not great right now either:
https://github.com/Frommi/miniz_oxide/pull/36
https://github.com/Frommi/miniz_oxide/pull/47 (probably also applies to the C backend)

I'm working with maintainers of both crates on improving them, but I cannot guarantee that they'll be safe after I'm done as I am NOT a security professional.

jonpas: Still looks better than libflate.

Also speed of flate2 is more than triple of libflate, unless that changes in libflate, flate2 is a clear winner.

malbarbo:
@mvdnes What is the status of this PR? It would be great if it could be merged. Many crates, like tectonic and reqwest already depends on flate2, so using zip crate together with one the two make the application uses two distinct deflate implementations.

Cargo.toml
Shnatsel: Latest release of miniz_oxide contains no unsafe code. It is now superior to libflate in every way.

mvdnes:
Since you made some compelling arguments, I have decided to merge this. Thanks for the contribution!

malbarbo:
Thank you!

Issue #16, opened 2016-08-24
Single-pass ZIP generation


tanriol:
`ZipWriter` currently expects the underlying stream to be seekable. The seekability is used to update the local file header after the file is complete (and also to determine some offsets).

The ZIP file format specification supports writing an archive in one pass without seeking. This mode may be useful in some cases (for example, streaming a ZIP archive of several large files without generating the archive ahead of time).

Would you be interested in implementing this mode and lifting the seekability requirement?

mvdnes:
I would love to have this, but lack the time and interest unfortunately.

It should not be to hard, as the only thing to do is to write the sizes and CRC after the data instead of before. The only real issue lies in determining the compressed file size, as that also uses the Seek trait at the moment.

gyscos:
It's probably fine to buffer each file individually (so we have the compressed size), but not the entire zip archive.

gyscos:
When writing the file, we can remember how much data was written to compute the compressed size.
When reading, it's probably fine to require the `Seek` trait as we need to jump between the files and the central directory anyway.

tanriol:
Technically single-pass reading is possible too as most information from the central directory is duplicated in the file headers too, IIRC.

NobodyXu:
Pinging this issue as I really want this feature!

Here in [`cargo-binstall`](https://github.com/ryankurte/cargo-binstall/issues/170), we are forced to write the downloaded file first to the disk, then pass it to `ZipArchiver` whereas all the other algorithms we supported (gzip, xzip and zstd) have a streaming interface that just accepts `BufRead` (and `Read` for convenience).

PR #114, opened 2019-10-05
Support for nonseekable Writer

izderadicka:
This is my initial attempt for #16 implementation.  
Although the actual change is not big: 
a) if stream is not seekable set 4th bit in general  flags in file local header
b) when file is finished do not update file local header, but instead write data descriptor record.

However significant work was need to make ZipWriter to work with both Seek and non Seek streams and also keep current interface and functionality untouched.  This should be non-breaking change - all current functionality works as it is. Only there is new associated method to create ZipWriter::new_streaming , which accepts steam that does not support Seek and through it new functionality will be available.

For this to work I had to do bit of type acrobatics. If somebody thinks it can be done in any other more easy,  straightforward way (and not to break current interface), please let me know. Main downside is that there is couple of supporting types, that must public as ZipWriter depends on them. I think it's similar to various Iterator types, that are very rarely used.

I made one basic test and also tested result with unzip utility in linux, where it worked.  I guess more tests and more work is needed, just wanted to check if this can be a way forward for this functionality.


Plecra:
This is very nice, thankyou! I don't see any reason not to merge this, but it's got a couple blockers: Could you rebase it on master and run `rustfmt` over your changes?

izderadicka:
@Plecra What about this - I see that build on 1.24 is failing but it is on miniz_oxide v0.4.0  which is dependency in code not used by this change.

Plecra:
Yes, it's unfortunately blocking the build at the moment. I should be able to manually review this soon

Plecra:
@izderadicka I'll review these changes within the next week, and hopefully add them to an upcoming release. Lmk if there are any more changes you'd like to make first 😄

nazar-pc:
Any updates on when maintainers will have time to upstream this?

nazar-pc:
Just a quick update, currently using code from this PR and it seems to work really well. Thanks Ivan!

Plecra:
Annoyingly, there are too many changes to master to merge this right now. I'd like to revisit the feature soon, and this PR proves we can do that in 0.6 without breaking changes. Thankyou for the contribution 😄

PR #115, opened 2019-10-19
Add ZipCrypto reading support

### LINKED ISSUE #64: Open ZIP archive protected by password

BenjaminRi:
Added two new functions to the API ( resolves #64 ):

```
ZipArchive::by_name_decrypt
ZipArchive::by_index_decrypt
```

These functions behave identically to their already existing counterparts without `_decrypt`, but take a password in the form of `&[u8]`. The password is not a `&str` because the ZipCrypto standard does not define which encoding is used, so UTF-8 cannot represent all possible passwords (for details, please read the comments in zipcrypto.rs).

The functions integrate seamlessly with the existing API. Substantial restructuring was necessary to insert the CryptoReader between the other readers. The code has been modified to facilitate the addition of further crypto or decompression routines.

Note that the API for `read_zipfile_from_stream` does not yet support ZipCrypto reading. However, this is not hard to add. Also, ZipCrypto writing support has not been added at this point in time. I want to make sure that my work is in line with what the authors and the users expect before I continue working on this.

In addition to that, a ZipCrypto writer will need cryptographically secure randomness (even though the crypto is weak). Where this notoriously hard to get randomness comes from is up for discussion.


BenjaminRi:
@mvdnes Any thoughts on this PR? I'm willing to invest time and fix issues if needed.

BenjaminRi:
Merged improvements from `master` into `pkzip-cipher` feature branch. Added unit test to verify correct ZipCrypto reader implementation. I have been using the feature branch regularly for half a year, and no problems were found so far. If there is interest to have ZipCrypto on `master`, I'm still open for feedback and willing to invest time.

Plecra:
This is brilliant! I hope we can get this merged soon. Would it be possible to support Central Directory Encryption with this implementation?

@rylev and I have recently been given permission to accept PRs, and should be able to give this a proper review soon.

BenjaminRi:
> Would it be possible to support Central Directory Encryption with this implementation?

Technically, yes. The ZipCrypto algorithm stands for itself and can be used to decrypt any binary stream, operating on the std::io::Read trait* (I also considered creating a separate crate just for ZipCrypto, but that's probably overkill). However, I am not sure if there is any ZIP software in existence that uses the weak ZipCrypto cipher to encrypt the Central Directory (relevant: https://github.com/nmoinvaz/minizip/issues/141 ).

A higher value task would be to support ZIP files with strong crypto like AES encryption which is believed to provide real security. It seems like Central Directory Encryption is often (always?) done with Strong Encryption. The encryption method of choice is almost always AES because all the alternatives in the specification are deprecated & insecure by now (RC2, RC4, DES, 3DES). AES encrypted ZIP files are somewhat common.

However, I think Strong Encryption should be treated in a separate issue/PR because it's a separate subject that requires its own considerations. The restructuring of the reader that I've done in this PR can be used to hook any other crypto algorithm into the reader pipeline, so it's a good preparation for further work.

*Also, encryption is almost finished, it just needs to get cryptographically secure randomness from somewhere.

BenjaminRi:
> LGTM! Could you run a `cargo fmt` on the changes?

Thanks! Do you mean just running `cargo fmt` on the changes or also additionally merging my branch such that it is up to date with master (26 commits happened in the meantime)?

BenjaminRi:
@Plecra Merged all the improvements from master. Applied `cargo fmt`.

BenjaminRi:
> Ah, sorry to be a pain, but I don't want to erase the commit history - I'd rather rebase the changes. I can handle that tomorrow if you'd like 😃

How does merging erase the commit history? The merge from `master` pulls in the entire commit history as well.

I used to rebase my changes, but for `read.rs` the process was so grueling that after a few failed attempts during a difficult rebase operation, I started to do merges only (maybe that is also an indication that this file needs to be split up). The merge in `read.rs` is not trivial. I'd appreciate it a lot if you could handle it if you think rebasing is the way to go. You should arrive at the same resulting source files as I did after the merge.

src/read.rs
```diff
@@ -223,26 +324,65 @@ impl<R: Read + io::Seek> ZipArchive<R> {
...
+    ) -> ZipResult<ZipFile<'a>> {
+        self.by_name_internal(name, Some(password))
+    }
+
     /// Search for a file entry by name
     pub fn by_name<'a>(&'a mut self, name: &str) -> ZipResult<ZipFile<'a>> {
+        self.by_name_internal(name, None)
+    }
+
+    fn by_name_internal<'a>(
```
rylev, line 341: I'd rather call this `by_name_with_optional_password`

src/read.rs
```diff
@@ -223,26 +324,65 @@ impl<R: Read + io::Seek> ZipArchive<R> {
...
+    fn by_index_internal<'a>(
+        &'a mut self,
+        file_number: usize,
+        mut password: Option<&[u8]>,
+    ) -> ZipResult<ZipFile<'a>> {
         if file_number >= self.files.len() {
             return Err(ZipError::FileNotFound);
         }
-        let data = &mut self.files[file_number];
+        let ref mut data = self.files[file_number];
```
rylev, line 377: I think we usually use the previous syntax. Is there a reason this was changed?

src/read.rs
rylev, line 341: I'd rather call this `by_name_with_optional_password`

src/read.rs
```diff
@@ -223,26 +324,65 @@ impl<R: Read + io::Seek> ZipArchive<R> {
...
 
-        if data.encrypted {
-            return unsupported_zip_error("Encrypted files are not supported");
+        if password == None {
+            if data.encrypted {
```
rylev, line 380: I think this would be easier to follow like this:
```rust
match (password, data.encrypted) {
  (None, true) => return Err(..),
  (Some(_), false) => password = None,
  _ => {}
}
```

src/zipcrypto.rs
```diff
@@ -0,0 +1,158 @@
+use std::num::Wrapping;
+
+struct ZipCryptoKeys {
+    key_0: Wrapping<u32>,
+    key_1: Wrapping<u32>,
+    key_2: Wrapping<u32>,
+}
+
+impl ZipCryptoKeys {
+    //Used this paper to implement ZipCrypto algo
```
rylev, line 10: Can this be part of the module documentation?

src/zipcrypto.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+pub struct ZipCryptoReader<R> {
+    file: R,
+    keys: ZipCryptoKeys,
+}
+
+impl<R: std::io::Read> ZipCryptoReader<R> {
+    pub fn new(file: R, password: &[u8]) -> ZipCryptoReader<R> {
+        //Note: The password is &[u8] and not &str because the documentation
```
rylev, line 58: nit: comments should have a space between the `//` and the first character

tests/zip_crypto.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    let mut archive = zip::ZipArchive::new(zip_file_bytes).unwrap();
+
+    assert_eq!(archive.len(), 1); //Only one file inside archive: `test.txt`
+
+    {
+        // No password
+        let file = archive.by_index(0);
+        assert!(file.is_err());
+        if let Err(error) = file {
```
rylev, line 50: You can match inside of the `Err` as well so there's no need for the match expression below

tests/zip_crypto.rs
```diff
@@ -0,0 +1,84 @@
...
+            match error {
+                zip::result::ZipError::PasswordRequired => (),
+                _ => panic!(),
```
rylev, line 53: Can we add some descriptive text to these panics?

tests/zip_crypto.rs
```diff
@@ -0,0 +1,84 @@
...
+            }
+        } else {
+            panic!();
+        }
+    }
+
+    {
+        // Wrong password
+        let file = archive.by_index_decrypt(0, "wrong password".as_bytes());
```
rylev, line 62: nit: you can use `b"wrong password"` instead of `as_bytes`

src/zipcrypto.rs
BenjaminRi: Fixed.

src/read.rs
BenjaminRi: Back when I wrote this code, I didn't even know this was possible with matcher clauses (I only learned that a week ago). Looks elegant, I applied that change.

src/read.rs
BenjaminRi: Okay. It's a bit long, but the name is clearer. Changed it to `by_name_with_optional_password` and `by_index_with_optional_password`.

src/read.rs
BenjaminRi: No, that was an oversight during the merge. Fixed.

tests/zip_crypto.rs
BenjaminRi: Fixed.

tests/zip_crypto.rs
BenjaminRi: Added some meaningful panic messages.

tests/zip_crypto.rs
BenjaminRi: Turned if expressions into cleaner matcher clause.

src/zipcrypto.rs
BenjaminRi: Improved & fixed docstrings throughout the entire `zipcrypto.rs` document. Added module level documentation.

BenjaminRi:
@rylev Thanks for the detailed review. I applied all your suggestions.

Plecra:
Thanks for following up on that! I've created #157 to discuss more encryption features if you'd like to share your thoughts 😉

BenjaminRi:
Fantastic! I'm very happy to see this PR accepted and merged 🎉

PR #116, opened 2019-10-27
Add function to extract a all files in an archive to a directory

zacps:
Adds a method `extract` to `ZipArchive` to extract all files in an archive to a directory. Based on the [extract example](https://github.com/mvdnes/zip-rs/blob/master/examples/extract.rs).

Fixes https://github.com/mvdnes/zip-rs/issues/62

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -6,8 +6,8 @@ use crate::result::{ZipError, ZipResult};
 use crate::spec;
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::io::{self, prelude::*};
 use std::fs;
+use std::io::{self, prelude::*};
 use std::path::Path;
 
 use crate::cp437::FromCp437;
--- a/src/read.rs
+++ b/src/read.rs
@@ -6,8 +6,9 @@ use crate::result::{ZipError, ZipResult};
 use crate::spec;
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::io;
-use std::io::prelude::*;
+use std::io::{self, prelude::*};
+use std::fs;
+use std::path::Path;
 
 use crate::cp437::FromCp437;
 use crate::types::{DateTime, System, ZipFileData};
@@ -231,6 +232,50 @@ impl<R: Read + io::Seek> ZipArchive<R> {
             offset: archive_offset,
             comment: footer.zip_file_comment,
         })
+    }
+
+    /// Extract a Zip archive into a directory.
+    ///
+    /// Paths are sanitized so that they cannot escape the given directory.
+    ///
+    /// # Platform-specific behaviour
+    ///
+    /// On unix systems permissions from the zip file are preserved, if they exist.
+    pub fn extract(&mut self, directory: &Path) -> ZipResult<()> {
+        for i in 0..self.len() {
+            let mut file = self.by_index(i)?;
+            let filepath = file.sanitized_name();
+
+            // `sanitized_name` should return a relative path
+            // otherwise there's a risk of directory traversal attacks
+            assert!(filepath.is_relative());
+
+            let outpath = directory.join(filepath);
+
+            if (&*file.name()).ends_with('/') {
+                fs::create_dir_all(&outpath)?;
+            } else {
+                if let Some(p) = outpath.parent() {
+                    if !p.exists() {
+                        fs::create_dir_all(&p)?;
+                    }
+                }
+                let mut outfile = fs::File::create(&outpath)?;
+                io::copy(&mut file, &mut outfile)?;
+            }
+
+            // Get and Set permissions
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::PermissionsExt;
+
+                if let Some(mode) = file.unix_mode() {
+                    fs::set_permissions(&outpath, fs::Permissions::from_mode(mode))?;
+                }
+            }
+        }
+
+        Ok(())
     }
 
     /// Number of files contained in this zip.
--- a/src/read.rs
+++ b/src/read.rs
@@ -238,6 +238,8 @@ impl<R: Read + io::Seek> ZipArchive<R> {
     ///
     /// Paths are sanitized so that they cannot escape the given directory.
     ///
+    /// This bails on the first error and does not attempt cleanup.
+    ///
     /// # Platform-specific behaviour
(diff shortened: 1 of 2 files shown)


tolvanea:
I agree this should be added. The amount of boilerplate code required for simple zip process is quite much. 

(If I would bother to send pull request, I would also add similar function to zip a directory.)

(Zip-rs is one of the most difficult libraries I have used. In my opinion, zipping/unzipping a directory should require only one line of code, but currently you need more than 50 LOC for that.)

src/read.rs
```diff
@@ -228,6 +230,50 @@ impl<R: Read+io::Seek> ZipArchive<R>
...
+    ///
+    /// On unix systems permissions from the zip file are preserved, if they exist.
+    pub fn extract(&mut self, directory: &Path) -> ZipResult<()> {
+        for i in 0..self.len() {
+            let mut file = self.by_index(i)?;
+            let filepath = file.sanitized_name();
+
+            // `sanitized_name` should return a relative path
+            // otherwise there's a risk of directory traversal attacks
+            assert!(filepath.is_relative());
```
JockeTF: Does this protect against someone using a path like `../../../alpaca.png`?

zacps: The [`sanitized_name`](https://docs.rs/zip/0.5.3/zip/read/struct.ZipFile.html#method.sanitized_name) function should do that. This is just an extra sanity check.

src/read.rs
```diff
@@ -228,6 +230,50 @@ impl<R: Read+io::Seek> ZipArchive<R>
...
+
+            let outpath = directory.join(filepath);
+
+            if (&*file.name()).ends_with('/') {
```
rylev, line 251: nit: is the `&*` necessary?

src/read.rs
```diff
@@ -228,6 +230,50 @@ impl<R: Read+io::Seek> ZipArchive<R>
...
     }
 
+    /// Extract a Zip archive into a directory.
+    ///
+    /// Paths are sanitized so that they cannot escape the given directory.
+    ///
+    /// # Platform-specific behaviour
+    ///
+    /// On unix systems permissions from the zip file are preserved, if they exist.
+    pub fn extract(&mut self, directory: &Path) -> ZipResult<()> {
```
rylev, line 240: I think we should follow the std lib and take `AsRef<Path>` instead of `&Path`. You can see an example [here](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.push).

src/read.rs
rylev, line 247: Should this be a `debug_assert`? Presumably if this ever gets tripped, it's a bug in zip-rs not in the user's crate. It feels wrong that we're forcing an extra runtime check to make sure this crate doesn't have bug.

src/read.rs
```diff
@@ -228,6 +230,50 @@ impl<R: Read+io::Seek> ZipArchive<R>
...
+            // `sanitized_name` should return a relative path
+            // otherwise there's a risk of directory traversal attacks
+            assert!(filepath.is_relative());
+
+            let outpath = directory.join(filepath);
+
+            if (&*file.name()).ends_with('/') {
+                fs::create_dir_all(&outpath)?;
+            } else {
+                if let Some(p) = outpath.parent() {
```
rylev, line 254: Can we use `matches!` instead of nested `if let` here? 
```rust
matches!(outpath.parent(), Some(p) if p.exists)
```

tests/extract.rs
```diff
@@ -0,0 +1,20 @@
...
+#[test]
+fn extract() {
+    let mut v = Vec::new();
+    v.extend_from_slice(include_bytes!("../tests/data/files_and_dirs.zip"));
+    let mut archive = ZipArchive::new(io::Cursor::new(v)).expect("couldn't open test zip file");
+
+    archive.extract(&PathBuf::from("test_directory")).expect("extract failed");
+
+    // Cleanup
+    fs::remove_dir_all("test_directory").expect("failed to remove extracted files");
```
rylev, line 19: This seems somewhat brittle, but I think we merge this without addressing this for now

src/read.rs
zacps: `matches!` is pretty recent (1.42.0, 2020-03-12), I'm not sure it's worth bumping the MSRV just for this?

src/read.rs
zacps: Looking at it now I'm not sure it's useful at all, removed.

tests/extract.rs
zacps: Agreed on both accounts, in future [tempdir](https://docs.rs/tempfile/3.1.0/tempfile/fn.tempdir.html) is probably the proper way to do this.

src/read.rs
rylev: Good point. We don't yet have a MSRV policy yet, but we should establish one

src/read.rs
```diff
@@ -233,6 +234,46 @@ impl<R: Read + io::Seek> ZipArchive<R> {
...
     }
 
+    /// Extract a Zip archive into a directory.
+    ///
+    /// Paths are sanitized so that they cannot escape the given directory.
+    ///
+    /// # Platform-specific behaviour
+    ///
+    /// On unix systems permissions from the zip file are preserved, if they exist.
+    pub fn extract(&mut self, directory: &dyn AsRef<Path>) -> ZipResult<()> {
```
rylev, line 244: Is there a reason that directory is a trait object instead of being a generic param?

src/read.rs
```diff
@@ -233,6 +234,46 @@ impl<R: Read + io::Seek> ZipArchive<R> {
         })
     }
 
+    /// Extract a Zip archive into a directory.
+    ///
+    /// Paths are sanitized so that they cannot escape the given directory.
```
rylev, line 239: Can we add a note that this will bail on the first failure and will not clean up after itself if it encounters an error?

src/read.rs
zacps: No particular reason. Slight decrease in compile time might be worth the cost of dynamic dispatch but it's really neither here nor there.

rylev: I personally find it surprising. I would personally like to have the policy of preferring static dispatch unless there's a reason not to. If we can change this, we can merge the PR

PR #117, opened 2019-11-11
Edition 2018

elpiel:


mvdnes:
Thank you for support. However, I am not a fan of running `cargo fmt` on the entire repository, as it breaks the history.

elpiel:
This is easily fixed, by reverting the last commit.
If you don't have anything against the other changes?

mvdnes:
Yes, I am fine with that.

PR #118, opened 2019-11-11
Fix fork

elpiel:


elpiel:
Oups, sorry, default branch is towards the forked repository.
We would like to try and test some impl for async version, we might eventually push it back here, but not sure how successful it will be.

PR #125, opened 2019-12-28
Update dependencies and bump version

Eijebong:


mvdnes:
The conversion to the new time library is not correct.
You have changed the tests at the end of `src/types.rs`  to `assert!(dt.to_time().is_err());`, but this should actually not produce an error. That it does now hints at a bug at other places, not in the test.

Eijebong:
Well the underlying time library can't represent these dates/times and just return an error when you try to build them

mvdnes:
O yes, you are right. I overlooked some of the parts of the date/times.

src/types.rs
```diff
@@ -106,24 +106,24 @@ impl DateTime {
...
-    pub fn from_time(tm: ::time::Tm) -> Result<DateTime, ()> {
-        if tm.tm_year >= 80 && tm.tm_year <= 207
-            && tm.tm_mon >= 1 && tm.tm_mon <= 31
-            && tm.tm_mday >= 1 && tm.tm_mday <= 31
-            && tm.tm_hour >= 0 && tm.tm_hour <= 23
-            && tm.tm_min >= 0 && tm.tm_min <= 59
-            && tm.tm_sec >= 0 && tm.tm_sec <= 60
+    pub fn from_time(tm: ::time::PrimitiveDateTime) -> Result<DateTime, ()> {
+        if tm.year() >= 80 && tm.year() <= 207
+            && tm.month() >= 1 && tm.month() <= 31
```
johannesmaibaum: The upper bounds check is wrong here, see also

johannesmaibaum:
```suggestion
            && tm.month() >= 1 && tm.month() <= 12
```

rylev:
@Eijebong this seems to be out of date and requires a rebase. Do you want to take a look at this again?

Plecra:
I'm not sure this is the right way to move forward with this. The `std::time` module has replaced much of the functionality of the original time crate.

Since the time fields in the zip spec are underspecified, it's going to be tricky to provide a robust API for them. The extended timestamp field is a very common zip extension, and maps 1-1 with `time::SystemTime` API. I think it's worth considering using that field, and filling `last mod file [date|time]` with a best-effort representation.

Is there something that the `time` crate provides us that justifies the extra dependency?

rylev:
Going to close this for now. Feel free to reopen the a PR or an issue, and we can revisit this.

PR #127, opened 2020-01-03
Fix month bounds check in `DateTime::from_time()`

johannesmaibaum:
The bounds checks for the `tm_mon` field of the `tm` parameter to `DateTime::from_time()` are wrong:
Instead of testing for range [0, 11], they are testing for [1, 31], probably due to a copy-paste error. Thus at the moment, *all otherwise valid dates in the month of January* will `Err` out.

To fix this:
* Fix bounds check of `tm.tm_mon` to [0, 11]
* Add test case to check the correct bounds for ZIP files as constrained by MS-DOS date and time bounds

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -333,6 +333,60 @@ mod test {
         assert!(DateTime::from_date_and_time(2107, 12, 32, 0, 0, 0).is_err());
     }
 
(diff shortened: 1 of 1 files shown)


mvdnes:
Thanks! :+1:

PR #130, opened 2020-01-14
Added ZipArchive::file_names() method to return all the filenames in archive.

### LINKED ISSUE #122: Implement of get a list of file names in zip archive
Consider if someone wants to get all the `png` files in an archive (I think this might be a common requirement), it seems that the only way for this at present is to iterate the whole archive with `.by_index()` and read the name of `ZipFile` and filter them. Maybe this can result in a decrease in performance? So can you imply a method for getting all the file names?
I just wonder, since there is a `names_map` field in struct `ZipArchive`, can it exposed to us (through some methods)? or this can result in some safety problems?

manuthambi:
Added ZipArchive::file_names() method to return an iterator of all the file names in the archive.

Using ZipArchive::by_index() to obtain a list of files is slow, if the files are
not read, because it creates the decompressor internally before returning ZipFile.

Fixes #122

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -256,6 +256,11 @@ impl<R: Read+io::Seek> ZipArchive<R>
         &self.comment
     }
 
+    /// Returns an iterator over all the file and directory names in this archive.
+    pub fn file_names(&self) -> impl Iterator<Item = &str> {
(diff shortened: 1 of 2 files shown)


mvdnes:
:+1:

PR #135, opened 2020-05-15
Use thiserror's Derive macro

stchris:
Hello,

I noticed that the existing codebase ~didn't compile~ issues warnings because `Error::description` is deprecated. I find that the macro from `thiserror` makes the `Error` enum more concise.

Edit: corrected the statement about the library not compiling, it just issues warnings

### DIFF
--- a/src/result.rs
+++ b/src/result.rs
@@ -1,94 +1,28 @@
 //! Error types that can be emitted from this library
 
-use std::convert;
-use std::error;
-use std::fmt;
 use std::io;
+
+use thiserror::Error;
 
 /// Generic result type with ZipError as its error variant
 pub type ZipResult<T> = Result<T, ZipError>;
 
 /// Error type for Zip
-#[derive(Debug)]
-pub enum ZipError
-{
+#[derive(Debug, Error)]
+pub enum ZipError {
     /// An Error caused by I/O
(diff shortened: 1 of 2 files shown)


src/result.rs
```diff
@@ -1,94 +1,28 @@
...
-{
+#[derive(Debug, Error)]
+pub enum ZipError {
     /// An Error caused by I/O
-    Io(io::Error),
+    #[error(transparent)]
+    Io(#[from] io::Error),
 
     /// This file is probably not a zip archive
+    #[error("Invalid Zip archive")]
```
rylev, line 18: nit: it might be a better idea to have the display strings start with lowercase letters. This is what `thiserror` shows in their documentation, and I think it generally makes for slightly nicer error messages.

stchris: ah, cheers for spotting that. I will fix it right away

stchris:
> I had one small nitpick. @stchris , let me know if you want to address that before merging, otherwise this looks good. 

Thanks for catching that, I updated the PR reflect those changes. 🙌 

> I can merge it as soon as @mvdnes gives me the rights

That sounds awesome! Thank you!

rylev:
Thanks a bunch!!! ❤

PR #136, opened 2020-05-15
Ensure crc32 checksum is checked only at the end of the reader

Dr-Emann:
The caller can pass in an empty slice for reading, which should return `Ok(0)`, not raise an error about an invalid checksum prematurely


mvdnes:
This was indeed an oversight. Thanks :+1:

PR #140, opened 2020-06-15
Move to GitHub Actions

rylev:
Since GitHub Actions is built in and supports all three OSes that we want to support, this is an easier maintenance burden.


rylev:
@mvdnes seems that there is an appveyor project configured to run on this. I don't have permissions to turn this off. Can you?

mvdnes:
I have removed the webhooks for AppVeyor and Travis

Issue #144, opened 2020-06-16
CRC checking in wasm is 150x slower


tonyhb:
When compiling to `wasm-unknown-unknown`, CRC checking slows decompression down by 150x.  Here's some profiling:

With CRC32, 8 seconds:

<img width="1914" alt="Screen Shot 2020-06-16 at 7 22 10 AM" src="https://user-images.githubusercontent.com/306177/84786930-5d764a80-afbb-11ea-8d1e-290bcfbe49db.png">

Without CRC32, 45ms:

<img width="1920" alt="Screen Shot 2020-06-16 at 7 20 53 AM" src="https://user-images.githubusercontent.com/306177/84786942-623afe80-afbb-11ea-9ba4-229624855087.png">

The allocations that CRC32 are making take seconds to finish, all within `crc32fast::baseline::update_fast_16`.  Within one call I count tens of allocations and copies.

I need to learn more about wasm to figure out how I can help, but wanted to put it out in the open to see if anyone had any ideas :)

Added a feature flag to cargo to test skipping CRC.  I don't feel like this is ideal, but if you want to test here's the commit: https://github.com/mvdnes/zip-rs/commit/f9f29cf7c1edebfa6489f9258b767bd4ad2a415e

rylev:
Hmm... skipping CRC doesn't sound like something we want to do. First thing, can you update the crcfast dependency to the latest version, just to make sure it's not an issue that has since been fixed?

tonyhb:
Yeah I definitely agree!  Only wanted to make sure that it was CRC checking and not the underlying flate performing allocations.  I'll take a look at upgrading now.  Thanks!

tonyhb:
@rylev confirmed that upgrading to crc32fast `1.2.0` does not fix the problem.  I'll report an issue over there.

tonyhb:
Turns out it's debug mode...!

PR #145, opened 2020-06-16
Add ability to skip CRC checking by feature flag

tonyhb:
Not taking this PR too seriously, but until I dive into why allocations within CRC32 checking are slow (#144) this alleviates the problem.  Obviously not ideal, you don't want to open a corrupt zip etc.

This is only intended to be used by the `wasm-unknown-unknown` target, and even then with caveats.


Plecra:
I assume this isn't important anymore since #144 closed

PR #148, opened 2020-06-16
Use the byteorder crate for Endianness handling

Plecra:
It's stable, and has 1713 (wow) direct reverse dependencies. Chances are, it's already part of users' dependency tree.


Cargo.toml
```diff
@@ -15,7 +15,7 @@ edition = "2018"
 [dependencies]
 flate2 = { version = "1.0", default-features = false, optional = true }
 time = { version = "0.1", optional = true }
-podio = "0.1"
+byteorder = "1.3.4"
```
rylev, line 18: Not: we are using “x.x” for dependencies. Any reason to use specifically version 1.3.4 instead of 1.3?

PR #150, opened 2020-06-17
Add deprecation warnings to potentially narrow the API

Plecra:
I believe we should attempt to gear the crate towards easily handling ZIP files correctly. There are enough warts introduced by the format itself, and it would be good to keep the API surface simple.

For now, deprecating these items lets users share how they're using them and gives us a chance to consider how to replace them. 

- Related to https://github.com/mvdnes/zip-rs/issues/58#issuecomment-644113649

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -34,6 +34,7 @@ impl CompressionMethod {
             8 => CompressionMethod::Deflated,
             #[cfg(feature = "bzip2")]
             12 => CompressionMethod::Bzip2,
+            #[allow(deprecated)]
             v => CompressionMethod::Unsupported(v),
         }
     }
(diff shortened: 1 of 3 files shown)


Plecra:
Hm... not sure how best to handle those warnings. Adding `allow` attributes everywhere is ugly, but we still want external deprecation warnings.

rylev:
Do you think it's worth going through the deprecation dance when the crate is still pre 1.0? I'm not sure what the previous policy was on breaking changes, but this strikes me as a being very conservative.

Plecra:
I don't feel particularly urgent about throwing these methods out, and it can wait for the rest of the crate to move to 0.6. I'd rather keep breaking changes to minor (0.x) versions whenever it's reasonable.

rylev:
This particular case seems small enough to where I'm not too bothered, but if we end up wanting to deprecate a lot more this could be a larger maintenance burden than it's really worth.

Plecra:
Agreed! The attributes would get very tedious with much more of this.

Issue #151, opened 2020-06-18
Error on Kobo while zipping .epub files


JeremyAube:
Using this library to zip .epub files results in an error when trying to open on Kobo Clara HD e-reader.

The error only occurs when the files are zipped with this library. When I zip files with system's `zip` command, the resulting .epub is valid.

Here is the comparison of the `zip-rs` file vs the system's `zip` command file using `zipinfo` on Ubuntu:
**zip-rs**
```
$ zipinfo zip-rs.epub 
Archive:  zip-rs.epub
Zip file size: 10702 bytes, number of entries: 9
-rw-r--r--  4.6 unx       20 b- stor 20-Jun-18 09:07 mimetype
-rw-r--r--  4.6 unx      253 b- defN 20-Jun-18 09:07 META-INF/container.xml
-rw-r--r--  4.6 unx      161 b- defN 20-Jun-18 09:07 META-INF/com.apple.ibooks.display-options.xml
-rw-r--r--  4.6 unx      563 b- defN 20-Jun-18 09:07 OEBPS/chapter_1.html
-rw-r--r--  4.6 unx     5687 b- defN 20-Jun-18 09:07 OEBPS/stylesheet.css
-rw-r--r--  4.6 unx     5758 b- defN 20-Jun-18 09:07 OEBPS/rust-logo.png
-rw-r--r--  4.6 unx     1109 b- defN 20-Jun-18 09:07 OEBPS/content.opf
-rw-r--r--  4.6 unx      497 b- defN 20-Jun-18 09:07 OEBPS/toc.ncx
-rw-r--r--  4.6 unx      645 b- defN 20-Jun-18 09:07 OEBPS/nav.xhtml
```

**zip command**
```
$ zipinfo zip.epub
Archive:  zip.epub
Zip file size: 10871 bytes, number of entries: 11
-rw-r--r--  3.0 unx       20 t- stor 20-Jun-18 08:42 mimetype
drwxrwxr-x  3.0 unx        0 b- stor 20-Jun-18 08:42 META-INF/
-rw-r--r--  3.0 unx      253 t- defN 20-Jun-18 08:42 META-INF/container.xml
-rw-r--r--  3.0 unx      161 t- defN 20-Jun-18 08:42 META-INF/com.apple.ibooks.display-options.xml
drwxrwxr-x  3.0 unx        0 b- stor 20-Jun-18 08:42 OEBPS/
[...]

PR #152, opened 2020-06-20
Add the text-file flag

mvdnes:
This adds the text flag which is stored in the internal file attributes.


mvdnes:
This is in response to the difference in the zipinfo output as seen in #151. This does not solve that bug however.

src/write.rs
```diff
@@ -133,6 +134,13 @@ impl<W: Write + io::Seek> Write for ZipWriter<W> {
                 "No file has been started",
             ));
         }
+
+        // Check if the buffer is text (UTF8) or contains binary data. Note that if an UTF8
+        // character is split between to calls to write, this will falsly mark it as binary.
+        // For ASCII characters, this is always correct.
```
rylev, line 140: Can we use `Utf8Error::valid_up_to` to figure out if we’re in this situation? I think to know for sure we would have to keep the bytes from any previous read the are potentially a non-complete character and if validation fails, prepend and check again.

src/types.rs
```diff
@@ -231,6 +231,8 @@ pub struct ZipFileData {
     pub header_start: u64,
     /// Specifies where the compressed data of the file starts
     pub data_start: u64,
+    /// Internal file attributes
+    pub internal_attributes: u16,
```
rylev, line 235: This is out of scope for this PR but I wonder if keeping this internally as a bitflags enum would be better....

src/write.rs
```diff
@@ -133,6 +134,13 @@ impl<W: Write + io::Seek> Write for ZipWriter<W> {
...
+        if self.stats.is_text == true && std::str::from_utf8(buf).is_err() {
+            self.stats.is_text = false;
+        }
```
Plecra, line 143: This behavior seems annoying - if the user has indicated that they're *trying* to create a text file, they probably don't want it to just silently fail. 

I'd like to investigate using `fmt::Write` to require the user to validate the input (`from_utf8_unchecked` when they want to be quick), but maybe there's a better solution.

For now, can we just return an error here instead? *something* is wrong with the input we've been given, and the user should know

Plecra:
I'd rather not provide this API like this - the issue came up because there are cases where it is significant whether or not this bit is set. There are cases where it will also be important that the flag *isn't* set, and this only leaves the user the option to deliberately provide invalid UTF-8 to get the flag set

src/write.rs
mvdnes: We could also just mark it as text for ASCII files. The appnote is a bit vague here:
```
       4.4.14.1 The lowest bit of this field indicates, if set, 
       that the file is apparently an ASCII or text file.  If not
       set, that the file apparently contains binary data.
```

mvdnes:
I am fine leaving this out for now. I do not know of any use cases which actually read this flag, except for `zipinfo`. This was a quick addition to test if this fixes the epub problem, but it did not seem to have any effect.

Plecra:
Thanks for bringing it up! We'll have to figure out a good solution in the end, but I don't think this is it - there are too many downsides. I'll leave the review to @rylev

rylev:
@mvdnes @Plecra do we know what the use cases are where utf-8 (or even just ascii) data should be marked as a binary file?

Plecra:
For now, we don't have any examples of code that would check the flag, no.

I'm fairly sure there'll be *some* zip-derived format that enforces the binary flag for, say, images, and images that technically contain valid unicode.

Also also, this will create false negatives too: "text" doesn't necessarily mean utf8.

rylev:
What seems most reasonable to me in the long run: offer a way for the user to decide how files are classified with helpers for such things as declaring all valid utf8 as text. The default would be ASCII being text and everything else being marked binary.

Plecra:
I'm moving this to issues as there haven't been any requests for the feature, and we're still not sure how best to implement it.

PR #155, opened 2020-06-21
Removed the default zip comment to an empty string

JeremyAube:
#151


src/write.rs
```diff
@@ -176,7 +176,7 @@ impl<W: Write + io::Seek> ZipWriter<W> {
             files: Vec::new(),
             stats: Default::default(),
             writing_to_file: false,
-            comment: "zip-rs".into(),
+            comment: "".into(),
```
Plecra, line 179: Could we use `String::new()` here? It makes the cheap operation clearer

JeremyAube:
I will change this to String::new() and I will also fix the test some other tests are failing because of the lack of comments

PR #159, opened 2020-06-23
Define the crate's MSRV

### LINKED ISSUE #158: Establish an MSRV (minimum supported Rust version)
We currently only run CI on the latest stable. We should establish a minimum version we want to support, ensure we run CI on that version and advertise that in the README.

Plecra:
This closes #158

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        rust: [stable]
+        rust: [stable, 1.34.0]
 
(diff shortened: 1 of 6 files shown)


Plecra:
Well, this would be that extra burden of dealing with the deprecated items 😅 

I can't consistently reproduce them yet, so I'm not sure what the cause is

Plecra:
Ah, the successful run had been cached. Looks like there was some change that allowed `#[allow(deprecated)]` in that position. Should we just go ahead and remove the deprecated items?

rylev:
I wouldn't mind... There's minor enough that we can get away with it. We should make note of it though in any change notes we do.

PR #175, opened 2020-07-08
chore: pin the flate2 version for now

Plecra:
This should fix the build until we have a more permanent solution. 

https://github.com/mvdnes/zip-rs/issues/170#issuecomment-655419242

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ Library to support the reading and writing of zip files.
 edition = "2018"
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -13,7 +13,7 @@ Library to support the reading and writing of zip files.
 edition = "2018"
 
 [dependencies]
-flate2 = { version = "1.0", default-features = false, optional = true }
+flate2 = { version = ">=1.0.0, <=1.0.14", default-features = false, optional = true }
```
rylev, line 16: Can we add a comment to the issue noting why we have such a strange version requirement?

PR #177, opened 2020-08-11
Fix overflow in directory counts of ZIP64 files

nickbabcock:
Cargo fuzz found an overflow corpus. I can't speak to any implications this has, but this patch fixes it

There is similar logic for non-zip64 files that may need to be examined more closely as well: https://github.com/mvdnes/zip-rs/blob/fbffaed42a0ca4393a12790c87897bd5fb9ad8f9/src/read.rs#L228


nickbabcock:
Been fuzzing all day and haven't been able to trip the other branch yet, so I'll defer to your preference on how you want this handled

Issue #180, opened 2020-08-19
Release 0.5.7


rylev:
We've made enough small non-breaking changes, that I think releasing 0.5.7 sounds like a good idea. @Plecra thoughts?

Plecra:
Agreed! I'd like to wrap up #178 first, and make a few additions to the documentation.

Would you mind adding that test? I imagine it should be fairly simple

Plecra:
Add #181 to that, which I think is quite urgent

robmv:
Just a small request. Can the release process include creating a tag on the repository? This is useful when you want to compare changes between releases, or to know what new fixes on master aren't still released.

Plecra:
I'd say we're ready to finalize the release. @mvdnes , how do you want to handle this?

mvdnes:
I see that we have a request for a tag. If you create one, I will publish that version to crates.io.

Plecra:
https://github.com/mvdnes/zip-rs/tree/v0.5.7

@rylev , that look good to you? If so, I'm happy for it to be published

rylev:
Let's do it! Thanks for handling this chore! For reference for anyone else, here's a [diff between 0.5.6 and 0.5.7](https://github.com/mvdnes/zip-rs/compare/18bece201c15ec78f5f46d51c1192aeb393742f2..v0.5.7) 

@mvdnes we're ready for you to pull the trigger! 😊

mvdnes:
Done :+1:

rylev:
Thanks!

@Plecra shall we start a tracking issue for 0.6?

Plecra:
Sure, it wouldn't hurt. We've got the milestone for now, but it would be good to explore the changes we want to make.

Briefly, (I'd say) we should be making iterative improvements on the current API, without many major changes to the internals.

BenjaminRi:
Thank you all for the good work and congratulations to @Plecra and @rylev for the first release since taking over.

Thanks to this new release, I was able to scrap my fork and depend on your zip crate, because you merged the ZipCrypto stuff.

Issue #181, opened 2020-08-19
Implement path sanitization


Plecra:
This was a surprise...

https://github.com/mvdnes/zip-rs/blob/62dc40678e2d90cccc1e583916b8cba47a6f3cb5/src/write.rs#L605-L616

This ignores all `Component::ParentDir`s, which could create very odd behavior in CLIs:

```SH
> cat bar
abc
> zip ./foo/../bar -o out
> unzip out into
> cat into/foo/bar # `bar` just moved into the `foo` directory! Wha?
abc
```

## Solution

Imo, this is quite clearly a logic bug, and should simply be remedied by including `ParentDir`s. However, this is the behavior documented for [`ZipWriter::start_file_from_path`](https://docs.rs/zip/0.5.6/zip/write/struct.ZipWriter.html#method.start_file_from_path), so we might not want to do that...

I'm tempted to simply deprecate the method, and warn users about the strange behavior. The implementation also uses `to_string_lossy`, which I don't think is really appropriate.

Plecra:
API that needs to be updated:

https://github.com/mvdnes/zip-rs/blob/0b46263eac1541d885dcee65582e2726056acd1a/tests/extract.rs#L8-L11
https://github.com/mvdnes/zip-rs/blob/0b46263eac1541d885dcee65582e2726056acd1a/src/read.rs#L325-L328
https://github.com/mvdnes/zip-rs/blob/0b46263eac1541d885dcee65582e2726056acd1a/src/read.rs#L612-L619

jamesmcm:
Deprecating this breaks CI pipelines that deny warnings. 

For now I worked around it by adding `#[allow(deprecated)]` but in general I think it'd be best to only deprecate when the deprecation message can have clear instructions of what is best to use instead. Since the vast majority of zip files won't be affected by this issue (despite it being an important issue).

Plecra:
Though it seems strange, I wouldn't mind adding a `mangled_name` method to direct people to. Would that do the trick @jamesmcm ?

jamesmcm:
Yeah, I think it's a good solution for now. Mainly since most zip files won't be affected.

PR #182, opened 2020-08-19
Preparing for 0.5.7

Plecra:
I've made a lot of minor changes to make the documentation more useful to a beginner. There's more to do, but it's a first step

As I was looking through it, I found that our path sanitization logic appears to be broken by design. It's behaving as documented, but would never be the correct behavior for a program. I've deprecated the methods, and had to (temporarily) remove the `ZipArchive::extract` method since it used the same logic.

This is related to #180


src/lib.rs
```diff
@@ -1,4 +1,7 @@
-//! A basic ZipReader/Writer crate
+//! An ergonomic API for reading and writing ZIP files.
+//!
+//! The current implementation is based on [PKWARE's APPNOTE.TXT v6.3.9](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT)
+// TODO: Decide on the crate's bias: Do we prioritise permissiveness/correctness/speed/ergonomics?
```
rylev, line 4: Do we need to include this TODO in the code? This seems like it would be better as an issue

src/read.rs
```diff
@@ -31,27 +31,24 @@ mod ffi {
...
 /// ```no_run
-/// use std::io::prelude::*;
-/// fn main() -> zip::result::ZipResult<()> {
+/// # use std::io::prelude::*;
+/// # fn main() -> zip::result::ZipResult<()> {
+/// # let buf: &[u8] = &[0u8; 128];
+/// # let mut reader = std::io::Cursor::new(buf);
+/// # let mut stdout = std::io::stdout();
+/// # let mut stdout = stdout.lock();
+/// let mut zip = zip::ZipArchive::new(reader)?;
```
rylev, line 43: I personally don't like examples where the user can't easily copy/paste the code and have it run. It can be confusing to new-comers what the `reader` and `buf` actually are.

src/read.rs
```diff
@@ -617,6 +611,11 @@ impl<'a> ZipFile<'a> {
 
     /// Get the name of the file in a sanitized form. It truncates the name to the first NULL byte,
     /// removes a leading '/' and removes '..' parts.
+    #[deprecated(
+        since = "0.5.7",
+        note = "by stripping `..`s from the path, the meaning of paths can change.
+                You must use a sanitization strategy that's appropriate for your input"
+    )]
     pub fn sanitized_name(&self) -> ::std::path::PathBuf {
```
rylev, line 619: Are we sure this is not a fixable bug? If it is fixable we shouldn't deprecate, we should just fix the bug 😄 I realize that would be a breaking change, but as we're pre-1.0, that's fine.

tests/extract.rs
```diff
@@ -2,20 +2,20 @@ extern crate zip;
 
 use std::fs;
 use std::io;
-use std::path::PathBuf;
 
 use zip::ZipArchive;
 
 // This tests extracting the contents of a zip file
 #[test]
+#[ignore]
```
rylev, line 10: Let's create an issue for the file sanitation problem and link to it here so we don't forget why this test is being ignored.

rylev (changes requested):
I had a few questions before we merge this

src/read.rs
Plecra: I definitely took this approach in the spirit of getting 0.5.7 out the door. There's no reason this couldn't be fixed in advance 👌

src/read.rs
Plecra: This *is* a no_run test - the uncommented code would always crash. We'd need to provide valid zip data to complete the example, which seems like it'd be distracting

src/lib.rs
Plecra: It'd be good to create an issue, yea. I'd like to keep the TODO though, since it's relevant to expanding the crate description, and would be useful to someone looking at that section of the code.

We could add a reference to the GH issue on this line?

tests/extract.rs
Plecra: Here we go https://github.com/mvdnes/zip-rs/issues/181

src/read.rs
mvdnes: This was not a bug but a feature. I tried to do the same as to do the same as python's [ZipFile.extract](https://docs.python.org/3.8/library/zipfile.html#zipfile.ZipFile.extract). Using this as the default will hopefully prevent many path-traversal bugs/exploits.

Plecra: I wouldn't consider python the best guideline for correct APIs 😉. I can't think of a situation where this behaviour would be desirable.

Here's a simple example of where it goes wrong:

```sh
> # Bob wants to share the source code of his Rust $MODULE
> DIR=${find . -type d -name $MODULE}
> zip source.zip $DIR $DIR/../$MODULE.rs
adding: ./src/foo/foo.rs
adding: ./src/foo.rs
```

```rs
ZipArchive::open(File::open("source.zip")?)?.extract(current_dir().join("out"))?;
```

```sh
> find out
out/src/foo.rs
# Where's `out/foo.rs`?
```

What's wrong with failing on paths that can't be normalised?

mvdnes: If you normalise path names this can also occur. You could add `x` and `a/../x`, which in my eyes would be equally undesirable. I admit that the current one is too.

The only truly safe way is to deny automatic extraction of files with a `..` in them.

Please note that it also removes `Prefix` and `RootDir` components, as specified in [std::path::Component](https://doc.rust-lang.org/std/path/enum.Component.html).

Plecra: I agree! There isn't a single correct way to handle these scenarios, which was the intention behind my "You must use a sanitization strategy that's appropriate for your input" comment. Maybe we should find a better wording.

Plecra: Then we'll deprecate for 0.5.7, with a plan to fix the API in 0.6. It's being tracked in #181

tests/extract.rs
Plecra: It doesn't seem appropriate to leave the disabled API in the release code, so I've just removed it for now. We can bring it back when we fix path sanitization in 0.6

Plecra:
Do we want to use release branches? It might make sense to use one for rolling back that `extract` API...

src/read.rs
Plecra: I'd add to that that `buf` isn't visible in the test. All the users need to recognize is that `reader` is referring to *any* `impl Read`, which I think is fairly reasonable. We could change this to a function that takes an `impl Read` itself, but that seems unnecessary to me

Plecra: I'd add to that that `buf` isn't visible in the test. All the users need to recognize is that `reader` is referring to *any* `impl Read`, which I think is fairly reasonable. We could change this to a function that takes an `impl Read` itself, but that seems unnecessary to me

rylev: Again, I don't like examples that aren't directly copy/pastable, but if you feel strongly that this is an improvement, we can leave it.

rylev:
@Plecra if you think that makes things easier. I honestly would have skipped the deprecation step and just released a fixed version as 0.6, but I recognize that that is a bit more aggressive deprecation policy than you would like.

src/read.rs
Plecra: I'm not clear what you consider better. Would you like the function signature to be uncommented?

Plecra: I'm not clear what you consider better. Would you like the function signature to be uncommented?

Plecra:
Deprecation warnings are the best way I know to discourage 0.5 users from using the API. They should be given a warning that their code is probably not behaving the way they expect without needing to accept breaking changes.

src/read.rs
rylev: I would prefer that yes. Or provide a concrete `Read` type as a local variable.

rylev (approved):
@Plecra shall we merge this?

PR #187, opened 2020-08-25
WIP: Update miniz_oxide to 0.4

### LINKED ISSUE #170: miniz_oxide 0.4.0 breaks 1.34.0 MSRV
Upstream issue: https://github.com/Frommi/miniz_oxide/issues/85

Plecra:
Closes #170 

The `no_extern_crate_alloc` has been published! Woop 🎉 

Rust dependency features can get weird, but I'm pretty sure this does the trick. `flate2` should compile with the latest version that can build with `miniz_oxide` 0.4.x, which (assuming they play nice with semver) should always have a `no_extern_crate_alloc` feature.

Edit: ack... `flate2` is going to need no-alloc support. Might still be waiting

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -11,12 +11,97 @@ use std::io;
 use std::io::prelude::*;
 use std::mem;
 
-#[cfg(any(
-    feature = "deflate",
-    feature = "deflate-miniz",
-    feature = "deflate-zlib"
-))]
-use flate2::write::DeflateEncoder;
+#[cfg(feature = "deflate")]
+mod deflate {
+    use miniz_oxide::deflate::core::{
+        compress_to_output, CompressorOxide, TDEFLFlush, TDEFLStatus as Status,
+    };
+    use miniz_oxide::deflate::CompressionLevel;
+    use std::io;
+
+    pub struct Encoder<W> {
+        inner: W,
+        compressor: Box<CompressorOxide>,
+    }
+    impl<W: io::Write> Encoder<W> {
+        pub fn new(inner: W, level: CompressionLevel) -> Self {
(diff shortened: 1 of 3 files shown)


rylev:
@Plecra can you edit this to be a WIP pull request until we figure out what to do with `flate2`?

Plecra:
I'm experimenting with depending directly on `miniz_oxide` atm, which would mean just dropping `flate2`. It's looking doable so far

Plecra:
We need to add and test the error handling in the deflate implementation, but I think this shows that maintaining the implementation ourselves gives us much more flexibility in the implementation (Maybe we can remove that `BufReader` for `io::BufRead`s?)

src/write.rs
```diff
@@ -11,25 +11,106 @@ use std::io;
...
+                match compress_to_output(compressor, buf, TDEFLFlush::None, |mut compressed| loop {
+                    match inner.write(compressed) {
+                        Ok(0) => return true,
+                        Ok(used) => compressed = &compressed[used..],
+                        Err(e) => {
+                            err = Some(e);
+                            return false;
+                        }
+                    }
+                }) {
```
Plecra, line 52: This isn't done yet, I was just experimenting with the error handling strategies. Afaik, we can't rewind the compressor, so *something* needs to be done with that `compressed` buffer. We're not allowed to return an error since we may have already written to `inner`, so we'll *probably* need to use an internal buffer to save results here. Maybe we reuse the buffer in the `compressor`?

zamazan4ik:
@Plecra do we still need this PR?

Plecra:
Nope! `flate2` has made the changes we needed upstream, so zip is okay for now. The door is still open to drop the `flate2` dependency since we aren't using much of their API beyond miniz

Issue #186, opened 2020-08-25
Add support for extra field


inorick:
Support for the [extra field](https://en.wikipedia.org/wiki/Zip_(file_format)#Extra_field) of ZIP file headers is currently marked as [future work](https://github.com/mvdnes/zip-rs/blob/8a6ca63728e4fe15e57e88efcc6c865d5e23680a/src/write.rs#L601).

My idea for an implementation is to add `*_with_extra_data` versions of the `start_file` and `start_file_from_path` methods that accept a slice with extra data to fill the extra field. The data would then be part of the `ZipFileData` struct.

I started preliminary work on it here: https://github.com/inorick/zip-rs/pull/1/commits/4393671828d2ab6ff95150d17f5109b2c05cc237

Please let me know what you think of this approach.

Plecra:
Thanks for starting on the design! This seems reasonable with the crate's current API, and I'd be happy to merge a simple implementation so long as it had a solid set of tests.

I'd only add that the user should be able to pass owned data to to `*_with_extra_data` as they do for the `name`. It's likely the buffers they create will be unique to each file.

Plecra:
I'd like to welcome anyone looking at this issue to make a PR implementing the `extra_field` API for 0.5. Though it would be very useful, I'm not planning on adding more features to this version of the API myself

n3vu0r:
I'm interested in aligned files which seems to be usually achieved by appending zeros to the local header as part of the extra field. Instead of passing the extra data directly, I would suggest to wrap it in a closure which itself provides a preliminary `data_start()`: 

```rust
pub fn start_file_with_extra_data<S, E>(&mut self, name: S, mut options: FileOptions, extra_data: E) -> ZipResult<()>
        where
            S: Into<String>,
            E: FnOnce(u64) -> AsRef<[u8]>,
```
This would allow to calculate the pad length which I think is not reliable otherwise:
```rust
zip.start_file_with_extra_data("file.ext", fopts, |data_start| vec![0; (align - data_start % align) % align]);
```
And it doesn't bloat the original use case:
```rust
zip.start_file_with_extra_data("file.ext", fopts, |_| extra_data);
```
What do you think? If nobody is working on a PR, I could give it a try soon.

n3vu0r:
Alternative design:

[...]

PR #188, opened 2020-09-02
support parsing Zip Unicode path extra field

vi117:
For backward compatibility, some zip archieve put the name encoding utf-8 in the Zip extra field.
This PR support parsing the extra field for [4.6.9 -Info-ZIP Unicode Path Extra Field](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT)


Plecra:
Thanks for the PR! We're working on exposing the extra field in #186 , which will let users parse the extensions they care about.

Why should we support this extension in particular? Afaict, there's nothing that sets it apart from the long list of others that we don't have the bandwidth to maintain.

vi117:
That seems to be correct. I close this PR.

ArcticLampyrid:
> Why should we support this extension in particular? Afaict, there's nothing that sets it apart from the long list of others that we don't have the bandwidth to maintain. (@Plecra )

I believe we should support this extension in particular. Since most tools on Windows store Unicode path in "Zip Unicode path extra field", any users should handle this extension properly otherwise they will face failure when handling non-ASCII char. It is especially common for non-English language to use Unicode chars.
Filenames are so basic an attribute that it deserves handing it at library level.


According to the spec:
> If both the File Name and Comment fields are UTF-8, the new General Purpose Bit Flag, bit 11 (Language encoding flag (EFS)), can be used to indicate that both the header File Name and Comment fields are UTF-8 and, in this case, the Unicode Path and Unicode Comment extra fields are not needed and SHOULD NOT be created.
> 
> Note that, for backward compatibility, bit 11 SHOULD only be used if the native character set of the paths and comments being zipped up are already in UTF-8.

For Windows users, the native character set is NEVER UTF-8. So any tool that meet the recommendation requires us to handle this extension properly.

PR #190, opened 2020-09-07
use Option to indicate missing file

### LINKED ISSUE #143: Remove `FileNotFound` error
Currently the `FileNotFound` error does not really represent an error case so much as the user has asked for a file that doesn't exist. This is normally represented as a `None` in an option. I think we should return `ZipResult<Option<ZipFile>>` from `ZipArchive::by_index` and `ZipArchive::by_name`

Plecra:
This is an implementation of, and would close #143 .

The extra `unwrap`s are a little jankier, but probably for the best.

*Note: This is decidedly a breaking change, which we should only make if we are now preparing for 0.6*

### DIFF
--- a/examples/extract.rs
+++ b/examples/extract.rs
@@ -17,7 +17,7 @@ fn real_main() -> i32 {
     let mut archive = zip::ZipArchive::new(file).unwrap();
 
     for i in 0..archive.len() {
-        let mut file = archive.by_index(i).unwrap();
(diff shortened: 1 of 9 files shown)


rylev (approved):
👍

zamazan4ik:
@Plecra am I right that this PR is still actual can just need to be rebased?

Plecra:
The changes are trivial, and need to be updated for the new APIs. It's not urgent, so no we'll close this for now.

PR #197, opened 2020-09-10
Remove breaking changes in 0.5.7

### LINKED ISSUE #193: 0.5.7 breaking change: Previously exhaustive ZipError matches no longer exhaustive
I have code mapping between error types which is no longer exhaustive:
```rust
match imp.writer.add_directory(path, FileOptions::default()) {
    Err(ZipError::FileNotFound)             => return Err(VfsError::FileNotFound { path: path.into() }),
    Err(ZipError::Io(e))                    => return Err(VfsError::IoError(e)),
    Err(ZipError::InvalidArchive(e))        => return Err(VfsError::IoError(io::Error::new(io::ErrorKind::InvalidData, e))),
    Err(ZipError::UnsupportedArchive(e))    => return Err(VfsError::Other { message: e.into() }),
    Ok(())                                  => {},
}
```
```
    |
46  | match imp.writer.add_directory(path, FileOptions::default()) {
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ patterns `Err(PasswordRequired)` and `Err(InvalidPassword)` not covered
    | 
```

It might be worth marking ZipError [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html) going forward, or adding a `#[doc(hidden)] _DoNotFullyMatch` variant to ZipError to avoid similar breakages in the future.

Plecra:
This should close #193 

It's not pretty, but at least this way we get to expose the crypto functionality in 0.5


rylev (approved):
Sigh... I really don't like this because it's clearly a bad API, but I guess we don't have a choice. We should be more careful in the future.

BenjaminRi:
@Plecra I strongly dislike this solution. I will be going back to my fork and wait for a 0.6 release, in which the API is hopefully restored to how it was with clean return codes.

Plecra:
@BenjaminRi Do you dislike the way we're having to use `UnsupportedArchive`? Would it solve the problem if we added a const to match with?
```rust
impl ZipError {
    const PASSWORD_REQUIRED: Self;
}
```

...admittedly the fix was a little rushed. Is there a better solution that you can see?

BenjaminRi:
@Plecra Yes, that is one thing I dislike. It also obscures what is an error and what is success by nesting `Result`s. (You see this in `Ok(Err(...))`). And the interfaces `by_name` and `by_name_with_optional_password` are no longer uniform because their return types differ. My idea was that these functions are identical, it's just that one of them takes a password. (Ideally I would have used function overloading or an optional parameter, but Rust lacks both of these. Maybe the builder pattern could be used.) With `Ok(self.by_name_with_optional_password(name, None)?.unwrap())`, it's also not immediately obvious that these `unwrap` calls cannot lead to panics inside the library, though it seems to be safe to unwrap.

But it's okay. As 0.5.8 is already out, it is released like that. Don't worry about me, I can build my stuff no matter what is released on crates.io. I just hope that eventually, a release 0.6 will follow that reverts this commit and restores how the interface was originally. Such that I can use this crate, rather than my fork. Ideally, the ZipCrypto feature should have shipped in a minor release, not a patch.

Plecra:
Hey, I totally agree that this API is a shame. I'd like to hear what you'd prefer to help with the next API design.

In particular, I'd probably use the `Result<Result<File, InvalidPassword>, Error>` pattern again, since I think it represents the error cases better. [Tyler Neely's post](https://sled.rs/errors.html) outlines this quite nicely, but the gist is that I don't want people to have to handle `InvalidPassword`s when they're only reading the central directory. It'd be great to know what you find obscuring about this. Is it just the presence of the `Ok`? I find that quite illustrative: "The password was successfully checked".

BenjaminRi:
I would just undo this commit for the next release where breaking changes are allowed. I think it was fine the way it was.

> but the gist is that I don't want people to have to handle InvalidPasswords when they're only reading the central directory.

Why not? Considering that central directories could technically be password protected, it would be a good fit. If the crate gets more features, this could be a very realistic scenario. I think missing and invalid passwords belong into the core error enum because these errors are ubiquitous in the wild.

The comparison to `compare_and_swap` is not apt here. `compare_and_swap` is expected to "fail" with `CompareAndSwapError` during normal program operation. It is a completely normal occurrence that `compare_and_swap` has to be retried multiple times until it works. I'd even argue that the name `CompareAndSwapError` is misleading. It's not an error, just like it's not an error if you scrub a dirty surface and the dirt only loosens up after the second or third time you run the brush over the surface. It's a normal part of an activity that eventually succeeds. Therefore, in this case, I do see the point of separating the `sled::Error` from the `CompareAndSwapError`. 

 In contrast, a wrong password really is a hard error that has to be handled similarly to when a file isn't found: A new course of action needs to be taken, a simple retry is not enough.

I understand that it's a problem if we have a huge central error that combines all the possible errors, particularly when functions that couldn't possibly have most of these errors do in fact return this huge error type. I don't think this is currently a problem in the zip crate. If we face such a complex error landscape, we need to rethink the entire error handling strategy, possibly using trait objects or combining multiple error enums in a parent error enum to deal with the complexity.

Plecra:
I'm sorry to say that we also don't think a file being missing is a hard error, and are expecting to model it with a `Result<Option<File>>` 😛.

I tend to think of how the API translates to a UX. Some parsing errors are so problematic that the program can't reasonably recover. If, say, the central header describes a malformed file, I'd want to display an error to the user and stop processing the invalid data. But when a user has mistyped their password, I want to give them the chance to try again. Similarly, they can look for another file if the one they searched for isn't there.

PR #198, opened 2020-09-12
Reintroduce Path Sanitization

### LINKED ISSUE #181: Implement path sanitization

Plecra:
Closes #181

I've kept the old behaviour available through `mangled_name`, for users who actively decided on it. I took the opportunity to give a stronger warning in the documentation, and provide an alternative which I think is more suitable.

This also means we can bring back the `ZipArchive::extract` API, which I now see as effectively complete for 0.5.

Are these API changes enough @jamesmcm ?

(And since you've shown interest in the old API, I'd also like to know which strategy you prefer @aspen 😀 )

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -9,6 +9,7 @@ use crate::zipcrypto::ZipCryptoReaderValid;
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::io::{self, prelude::*};
+use std::path::Component;
 
 use crate::cp437::FromCp437;
 use crate::types::{DateTime, System, ZipFileData};
@@ -564,11 +565,21 @@ impl<'a> ZipFile<'a> {
     }
 
     /// Get the name of the file
+    ///
+    /// # Warnings
+    ///
+    /// It is dangerous to use this name directly when extracting an archive.
+    /// It may contain an absolute path (`/etc/shadow`), or break out of the
+    /// current directory (`../runtime`). Carelessly writing to these paths
+    /// allows an attacker to craft a ZIP archive that will overwrite critical
+    /// files.
     pub fn name(&self) -> &str {
         &self.data.file_name
     }
 
     /// Get the name of the file, in the raw (internal) byte representation.
+    ///
+    /// The encoding of this data is currently undefined.
     pub fn name_raw(&self) -> &[u8] {
         &self.data.file_name_raw
     }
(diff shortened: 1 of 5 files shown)


jamesmcm:
Thanks, this is great!

src/read.rs
```diff
@@ -310,6 +311,49 @@ impl<R: Read + io::Seek> ZipArchive<R> {
             comment: footer.zip_file_comment,
         })
     }
+    /// Extract a Zip archive into a directory.
+    ///
+    /// Malformed and malicious paths are rejected so that they cannot escape
```
rylev, line 316: This should probably say "Malformed or potentially malicious paths".

Perhaps we should make reference to `name_as_child` docs to enumerate the types of paths that will be rejected.

src/read.rs
```diff
@@ -310,6 +311,49 @@ impl<R: Read + io::Seek> ZipArchive<R> {
...
+
+        for i in 0..self.len() {
+            let mut file = self.by_index(i)?;
+            let filepath = file
+                .name_as_child()
+                .ok_or(ZipError::InvalidArchive("Invalid file path"))?;
+
+            let outpath = directory.as_ref().join(filepath);
+
+            if (file.name()).ends_with('/') {
```
rylev, line 335: Why the `()` around `file.name()`?

src/read.rs
```diff
@@ -310,6 +311,49 @@ impl<R: Read + io::Seek> ZipArchive<R> {
             comment: footer.zip_file_comment,
         })
     }
+    /// Extract a Zip archive into a directory.
+    ///
+    /// Malformed and malicious paths are rejected so that they cannot escape
+    /// the given directory.
+    ///
+    /// This bails on the first error and does not attempt cleanup.
```
rylev, line 319: Perhaps we should also mention that this will override existing files if they have the same path as a file in the zip archive.

src/read.rs
```diff
@@ -578,12 +635,55 @@ impl<'a> ZipFile<'a> {
...
+    ///
+    /// - It can't contain NULL bytes
+    /// - It can't resolve to a path outside the current directory
+    ///   > `foo/../bar` is fine, `foo/../../bar` is not.
+    /// - It can't be an absolute path
+    ///
+    /// This will read well-formed ZIP files correctly, and is resistant
+    /// to path-based exploits. It is recommended over
+    /// [`ZipFile::mangled_name`].
+    pub fn name_as_child(&self) -> Option<&Path> {
```
rylev, line 670: Not sure how I feel about the name of the method. Perhaps `relative_name` or `checked_relative_name`. Nothing stands out as clearly better.

rylev (changes requested):
Awesome! I have a few changes I'd like to see, and I'd like us to bikeshed the name a bit more, but overall this is great!

Plecra:
Nice points @rylev. I've renamed the API to `enclosed_name`. It's not perfect, but I don't think we'll come up with something better with more time. I'd like to release 0.5.9 (with this fix) within the next couple days, so lmk if there's an issue

PR #203, opened 2020-10-19
Support AES decryption

mbr:
This PR adds support for decrypting AES encrypted zip archives (compression method 99). It implements the special non-NIST variant of the CTR mode used by WinZip, but otherwise uses off-the-shelf crypto crates.

The feature itself is hidden behind an `aes` feature flag. We will leave it up to upstream to enable or disable it and/or mention it in the docs.

Encryption can be added if desired, as the decryption methods can be used to encryption as well.


ZaMaZaN4iK:
@mbr can you please fix the conflicts? :) Then, I hope, such useful feature will be merged.

mbr:
> @mbr can you please fix the conflicts? :) Then, I hope, such useful feature will be merged.

I think this might be a job for @Lireer :)

Lireer:
All conflicts have been resolved. The MSRV is now `1.42`, which shouldn't be much of a problem considering the current stable version is `1.54`. `1.41` would be possible too, if we remove the two uses of the `matches!` macro.

zamazan4ik:
@Lireer sorry for sooooo long time for the review. Do you want to add to this PR anything else? Maybe some benchmarks, tests? If not and you think that it's ready to use, I want to merge it.

zamazan4ik (approved):

zamazan4ik:
@mbr Could you please rebase your work on `master` branch? Maybe you have some valuable additions to your PR? Thanks in advance!

Lireer:
The branch is up-to-date with master and could be merged. I added some simple tests which test the different AES encryption schemes (128/192/256) and whether they work with compressed and uncompressed files.

Benchmarks would be quite interesting but would require some archive to be decrypted. That would mean either implementing encryption, but I don't know enough about cryptography to give any guarantees about security, or including a benchmark file.
Here some results of local benchmarks:

| encryption | compression | file size | MB/s |
| ---------- | ----------- | --------: | ---: |
| none       | none        |     1 MiB | 6744 |
| none       | deflate     |     1 MiB | 1340 |
| none       | bzip2       |     1 MiB |   15 |
| AES128     | none        |      1 MB |   37 |
| AES128     | deflate     |      1 MB |   42 |
| AES256     | none        |      1 MB |   29 |
| AES256     | deflate     |      1 MB |   33 |
| AES256     | none        |     10 MB |   29 |
| AES256     | deflate     |     10 MB |   33 |

Cargo.toml
```diff
@@ -11,11 +11,16 @@ Library to support the reading and writing of zip files.
 edition = "2018"
 
 [dependencies]
-flate2 = { version = "1.0.0", default-features = false, optional = true }
-time = { version = "0.3", features = ["formatting", "macros" ], optional = true }
+aes = { version = "0.6.0", optional = true }
```
agourlay, line 14: What about using the latest aes version [0.7.5](https://crates.io/crates/aes)?
There seems to be breaking changes in the API.

Thank you for the hard work!

tests/aes_encryption.rs
```diff
@@ -0,0 +1,78 @@
+use std::io::{self, Read};
+use zip::ZipArchive;
+
+const SECRET_CONTENT: &str = "Lorem ipsum dolor sit amet";
+
+const PASSWORD: &[u8] = b"helloworld";
```
agourlay, line 6: Are there restrictions on the characters possible in the password?
I am thinking upper cased letters, punctuation, digits and so on.

Lireer: `aA1!@#$%^&*()_+=-\|}]{[` works as a password. In general anything should be possible, since passwords are just a slice of bytes. I wanted to try a password with an emoji, but the hard part is encryption, since p7zip only supports ascii.

Cargo.toml
Lireer: I tried updating aes, pbkdf2, hmac and sha-1 (now sha1) together since they all use the same dependencies. Since this required some small changes to the decryption code, I also put in some more (possibly meaningless) checks.

This increased the performance of AES-128 from 42 MB/s to around 160 - 180 MB/s.

zamazan4ik:
Just approved for running the CI

Lireer:
The new commit changes the name of `sha-1` to `sha1`. Both point to the same crate, but `sha1` recently changed ownership to RustCrypto, see [here](https://github.com/RustCrypto/hashes/pull/350). `sha1` will continue to be developed and supported, while `sha-1` will be deprecated with the next minor version update.

zamazan4ik:
@Lireer please ping me or @Plecra when you will finish the work. By the way, maybe a more convenient way for discussing the PR will be our new Discord server - https://discord.gg/rQ7H9cSsF4 ? :)

Lireer:
@zamazan4ik I consider this PR finished in the sense that it does what it's supposed to do, add aes decryption support. Things like benchmarks, aes encryption, and possible performance improvements can be added in other PRs.

src/compression.rs
```diff
@@ -91,6 +96,7 @@ impl CompressionMethod {
             8 => CompressionMethod::Deflated,
             #[cfg(feature = "bzip2")]
             12 => CompressionMethod::Bzip2,
+            99 => CompressionMethod::Aes,
```
zamazan4ik, line 99: Should we mode it under the feature?

src/compression.rs
```diff
@@ -115,6 +121,7 @@ impl CompressionMethod {
             CompressionMethod::Deflated => 8,
             #[cfg(feature = "bzip2")]
             CompressionMethod::Bzip2 => 12,
+            CompressionMethod::Aes => 99,
```
zamazan4ik, line 124: Should we mode it under the feature?

zamazan4ik:
@Plecra if you have no additional comments (and I am really not sure regarding my comments - maybe it's intended design to not move these values under the feature), I want to merge this PR.

src/compression.rs
Lireer: Having the `Aes` variant outside the feature is intended. It doesn't require any dependencies and makes it possible to recognize encrypted files and give a more fitting error message, see [here](https://github.com/zip-rs/zip/pull/203/files#diff-f5613ff66dd29a0567e4173ff7c04ed28b08eb7384040946dd410698700b87eeR213-R218).

zamazan4ik:
LGTM

src/aes.rs
```diff
@@ -0,0 +1,196 @@
...
+            // The following call to `finalize` consumes `hmac` so we replace `self.hmac` with a
+            // dummy that uses a `Key` made up of only zeroes. `self.hmac` should not be used after
+            // this.
+            let hmac = std::mem::replace(
+                &mut self.hmac,
+                Hmac::new(GenericArray::from_slice(
+                    &vec![0; <Hmac<Sha1> as KeySizeUser>::KeySize::to_usize()],
+                )),
+            );
+            let computed_auth_code = &hmac.finalize().into_bytes()[0..AUTH_CODE_LENGTH];
```
Plecra, line 174: We can use `finalize_reset` here to make this simpler to read

src/aes_ctr.rs
```diff
@@ -0,0 +1,324 @@
...
+
+/// This trait allows using generic AES ciphers with different key sizes.
+pub trait AesCipher {
+    fn crypt_in_place(&mut self, target: &mut [u8]);
+}
+
+/// XORs a slice in place with another slice.
+#[inline]
+fn xor(dest: &mut [u8], src: &[u8]) {
+    debug_assert_eq!(dest.len(), src.len());
```
Plecra, line 143: Possible optimization here if we make this an assert_eq

src/aes_ctr.rs
```diff
@@ -0,0 +1,324 @@
...
+    fn new_with_wrong_key_size() {
+        AesCtrZipKeyStream::<Aes128>::new(&[1, 2, 3, 4, 5]);
+    }
+
+    // The data used in these tests was generated with p7zip without any compression.
+    // It's not possible to recreate the exact same data, since a random salt is used for encryption.
+    // `7z a -phelloworld -mem=AES256 -mx=0 aes256_40byte.zip 40byte_data.txt`
+    #[test]
+    fn crypt_aes_256_0_byte() {
+        let ciphertext = [];
```
Plecra, line 165: The common part of these tests should be in its own function. Probably `roundtrips(&[], b"", [0x0b, ...])`

src/compression.rs
Plecra: Same problem as #266. We'll have to change `CompressionMethod::Aes` to `CompressionMethod::AES`, but I dont mind doing that myself (sadly I can't edit the PR :D)

Plecra (approved):
LGTM. Thanks a ton for the work on this!

zamazan4ik:
@Lireer if you want - please resolve the comments from @Plecra in this PR. If not - I think we can do it in another PR. And thanks for the so significant work!

src/aes.rs
Lireer: Before I realized that the `Reset` trait is now feature-gated, this was the only way to make it work. After I realized that `finalize_reset` can still be used I left it like this, because I thought the `Hmac` containing the password would be dropped earlier like this. But I guess it probably doesn't improve security at all, so making it more readable is more important.

src/aes_ctr.rs
Lireer: Doesn't seem to affect performance all that much. I will try with criterion if that makes it more noticable. Bencher runs report 2130 - 2160 MB/s with `debug_assert_eq` and 2120 - 2130 MB/s with `assert_eq`. Note the over 2000 MB/s, which is substantially more than the previously reported 170 MB/s.  Differences to previous benchmarks are: Windows instead of NixOs and a newer CPU (probably the reason). The other benchmark runs about 50% faster too.

src/compression.rs
Lireer: The constant has been added, but not sure if it's how you want it to be used.

src/aes_ctr.rs
Lireer: Done

src/aes_ctr.rs
Lireer: Seems like badly written benchmark code was the real reason for the performance improvement. But the criterion benchmarks with high sample sizes don't show any significant changes when decrypting a `ZipFile`.

Lireer:
@zamazan4ik @Plecra thanks for the reviews, if you don't have any other comments the PR should be ready to be merged.

zamazan4ik (approved):

Plecra:
ack! my bad for not being clearer, but we do need to remove `CompressionMethod::Aes` completely. I'll make a new issue for it

PR #208, opened 2020-11-10
Add extra field support

inorick:
This change adds initial read and write support for extra fields of ZIP files.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -261,7 +261,7 @@ impl<R: Read + io::Seek> ZipArchive<R> {
                 let directory_start = footer
                     .central_directory_offset
                     .checked_add(archive_offset)
-                    .ok_or_else(|| {
+                    .ok_or({
                         ZipError::InvalidArchive("Invalid central directory size or offset")
                     })?;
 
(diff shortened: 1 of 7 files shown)


src/read.rs
```diff
@@ -349,7 +349,7 @@ impl<R: Read + io::Seek> ZipArchive<R> {
     }
 
     /// Search for a file entry by name
-    pub fn by_name<'a>(&'a mut self, name: &str) -> ZipResult<ZipFile<'a>> {
+    pub fn by_name(&mut self, name: &str) -> ZipResult<ZipFile> {
```
Plecra, line 352: This will unify the lifetimes of `self` and `name`, changing the API. Did clippy recommend this change?

Plecra:
Hiya! Thanks for the PR. Frustratingly, I haven't had the time to properly review PRs in the last couple months due to other commitments. I'm hoping to find a weekend to set aside for merging some of the changes soon.

It'd be great if you could explain how this differs from #200, which I'm almost ready to merge.

inorick:
I had a look at #200 and it looks like a more complete implementation that also handles ZIP64. I did not test it but from the looks of it I would prefer it.

inorick:
Closing in favour of #200

PR #209, opened 2020-11-10
Provide constants for valid variants of CompressionMethod

### LINKED ISSUE #205: API for reading raw file content
Hi all.
I'm using zip-rs to only read zip archives (i.e. not decompress the content), and thus have default-features = false.
However the two variants of the enum CompressionMethod, Deflated and Bzip2 only exist if respectively the two features "flate" and "bzip2" are turned on. Currently this is not a big problem if you want to know the compression method, since you can just match on the Unsupported(v) variant and check the value v, however I can see that you are deprecating this variant.

So what am I suppose to do if I want to know the compression method but not turn on the decompression/compression features? Second is there a good reason for the variants Deflated and Bzip2 to be behind a feature flag?

Otherwise thanks for a really useful and easy to use crate :)

Nicklas

Plecra:
This should close #205 . Will this do the trick @nicklaswj?
image

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -31,41 +31,41 @@ pub enum CompressionMethod {
 #[allow(deprecated, missing_docs)]
 /// All compression methods defined for the ZIP format
 impl CompressionMethod {
-    pub const STORE: Self = Self::Stored;
-    pub const SHRINK: Self = Self::Unsupported(1);
-    pub const REDUCE_1: Self = Self::Unsupported(2);
-    pub const REDUCE_2: Self = Self::Unsupported(3);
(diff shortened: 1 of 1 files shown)


nicklaswj:
Not totally, but because of another problem I discovered that is related to this one:
The function [make_reader](https://docs.rs/zip/0.5.8/src/zip/read.rs.html#136) returns an error if the compression is unsupported, which of course makes good sense, but since it get called on the construction of ZipFile, I am still unable to use this crate without enabling the compression features

n3vu0r:
Typo: `ZSTD_DPERECATED`.

PR #211, opened 2020-11-15
Support extra field

Plecra:
Continuation of #200

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -67,6 +67,8 @@ pub struct ZipWriter<W: Write + io::Seek> {
     files: Vec<ZipFileData>,
     stats: ZipWriterStats,
     writing_to_file: bool,
+    writing_to_extra_field: bool,
+    writing_to_central_extra_field_only: bool,
     comment: String,
 }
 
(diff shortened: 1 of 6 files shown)


src/write.rs
```diff
@@ -466,7 +472,7 @@ impl<W: Write + io::Seek> ZipWriter<W> {
             options.permissions = Some(0o644);
         }
         *options.permissions.as_mut().unwrap() |= 0o100000;
-        self.start_entry(name, options)?;
+        self.start_entry(name, options, None)?;
```
n3vu0r, line 475: I think this change will switch the the writer too early resulting in compressed extra data. That's why I moved the switching out of `start_entry()` and did it explicitly.

Plecra: Great to know, thanks ^^ I'm expecting there to be a few conflicts with the other PRs that were hidden in the merge, and this will be good to keep in mind.

Btw if you'd rather keep #200 active, feel free to make this merge yourself. I just wasn't able to edit your PR

n3vu0r: Ok, I will keep #200 conflict-free and try to cover this case with a test. Enabling compression in `end_to_end.rs` should do it.

n3vu0r: Done, compression is enabled in the test by using `FileOptions::default()` for the file with extra data.

Plecra:
Closing in favor of #200

PR #213, opened 2020-12-08
Support symbolic links

### LINKED ISSUE #77: How to handle symlinks
Hello,

I have a ZIP archive created on linux using the `--symlinks` option.
So my archive contains symlinks.

Yet, when I extract the corresponding files, they end up being text files with the name of the symlink target as their content.

How can I detect/handle symlinks?

Regards,

markmmm:
This is to resolve #77 

This probably needs some re-work - but hopefully this is the correct direction.

I am definitely open to design changes/feedback.

Questions:
* Should we create symbolic links on Windows - or just make a copy of the file?
  * If yes - what is the best way to convert the `Vec<u8>` to a valid `&Path`
* I am unclear how to test (the other tests don't seem to actually extract - which would be required to test this). Maybe I need to restructure for testing?
* Do I have terminology of "source"/"target" consistent (i.e. should I rename `get_symlink_source()` to `get_symlink_target()`?)

I have disabled symlink creation on windows. I can't find/google a way create a `Path`/`OsStr` from a `Vec<u8>`. I was going through String first - but then was running into the issue that the conversion error is not handled as part of `ZipError` (and it also seems wrong anyway). 
1. Going through String seems wrong to me - as Windows like Linux can support non-UTF-8 paths (so code for Zip file handling shouldn't really expect that paths are valid UTF-8)
2. If we decide to use this (using String) for now then should we add a new error type to `ZipError`?

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -217,21 +217,14 @@ fn get_symlink_source(entry: &mut ZipFile<'_>) -> ZipResult<Option<Vec<u8>>> {
     }
     Ok(None)
 }
-fn create_link(link_target: Vec<u8>, link_path: &Path) -> ZipResult<()> {
-    #[cfg(target_family = "unix")]
-    {
-        // Needed to be able to call `OsString::from_vec(Vec<u8>)`
-        use std::os::unix::ffi::OsStringExt as _;
 
-        let link_target = std::ffi::OsString::from_vec(link_target);
-        std::os::unix::fs::symlink(link_target, link_path)?;
-    }
-    // #[cfg(target_family = "windows")]
-    // {
-    //     // TODO: Support non-UTF-8 paths (currently only works for paths which are valid UTF-8)
-    //     let link_target = String::from_utf8(link_target)?;
-    //     std::os::windows::fs::symlink_file(link_target, link_path)?;
-    // }
+#[cfg(target_family = "unix")]
+fn create_link(link_target: Vec<u8>, link_path: &Path) -> ZipResult<()> {
+    // Needed to be able to call `OsString::from_vec(Vec<u8>)`
+    use std::os::unix::ffi::OsStringExt as _;
+
+    let link_target = std::ffi::OsString::from_vec(link_target);
+    std::os::unix::fs::symlink(link_target, link_path)?;
 
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


cecton:
@Plecra ☝️  This PR is waiting for your feedback for some time 🙏  I can do a first pass review to help if you want.

Plecra:
Heya! Thanks for the offer, I've been struggling to keep up with maintenance for a while due to exam prep.

I'll quickly outline my thoughts here, though I may have missed something important while skimming: I'm hesitant to add an API without clear platform-independent behaviour. I would like to make it clear which specification we're using (it doesn't need to be super official! Just well known). It also seems strange to implement this without any symmetry in the writer API.

The main problem here is how ZIP symlinks are seemingly defined by the Info-ZIP implementation. I'd be much more interested in supporting an external crate that implemented the symlink extension (perhaps among others. We'll need to do something similar with the variants of timestamps once the extra field PR is merged). As far as I can tell, this should already be possible with the public APIs.

zamazan4ik:
@markmmm maybe you have updates or thoughts regarding the @Plecra comment?

zamazan4ik:
We decided to close this PR according to security reasons. We still need to implement symlinks support into the library but we need to investigate other libraries for symlink design decisions and then implement it here.

If you will have any suggestions for future design (like ideas, useful links, and so on) - feel free to comment here: https://github.com/zip-rs/zip/issues/77

markmmm:
Sorry for the delayed response. I really didn't know how to process @Plecra's comment :/
In the end - I created a wrapper around ZipArchive/ZipFile to do what I needed. I will see if I can update this PR with the latest code (for a reference - not for merging).

For example - I recently ran into a case where it failed on Windows due to Long path name restrictions - and I added a fix for that (and I guess that might be useful to someone else picking up this work?)

dvc94ch:
> security reasons

what security reasons are there exactly? I hope this gets fixed soon. for now I'll have to fallback to using `Command::new("unzip")` and hope every platform has that command installed.

markmmm:
>what security reasons are there exactly? I hope this gets fixed soon. for now I'll have to fallback to using Command::new("unzip") and hope every platform has that command installed.

Here is an explanation of 1 security issue - a zip could contain a symlink to a file outside of the zip structure...
https://security.stackexchange.com/questions/73718/how-zip-symlink-works

Some other links/info
* Python's `zipfile` [doesn't support symlinks](https://github.com/python/cpython/issues/82102).
  * Some support can be found in [zipfile2](https://github.com/cournape/zipfile2/blob/master/zipfile2/_zipfile.py)
* OpenBSD PR: Prevent an archive from esacaping the current directory by itself. https://github.com/openbsd/src/commit/6b45b4706c10e32905d6be347ff82fceb9bcf8f4
* Permissions issues (so need the ability to not extract symlinks: https://github.com/srikanth-lingala/zip4j/issues/278
* And more https://codeql.github.com/codeql-query-help/go/go-unsafe-unzip-symlink/  (with more links inside).
  * This is probably a good primer on what to do (and what to check)

Requirements:
* Allow extracting symlinks with an option (default would probably be to now extract symlinks)
* When not extracting symlinks - the output should be the contents of the linked file
* When extracting symlinks:
  * link should never 'escape' the extraction path
  * any symlink that points outside of the extraction folder would either be an error(?)  (need to check if target is also a symlink!)
  * .... more? ...

Issue #156, opened 2020-06-21
WIP: API Design


Plecra:
After sifting through the specification and the issues left in the repository, I've mocked up what I think would be a good API to work towards. It's incomplete, and I mostly bring it up as food for better ideas. Anyway, goals...

- Lossless modification: We want to be able to reliably modify archives we can't fully parse
  - [It's common for zip utilities to implement the format incorrectly](https://stackoverflow.com/questions/13261347/correctly-decoding-zip-entry-file-names-cp437-utf-8-or), and we want to be able to ignore information we're not touching
- In-place modification: Interpreting and editing the contents of an existing ZIP file
  - It's what the format was originally designed to be good at
  - ...but also less useful for today's purposes. I just think this would be fun to pull off.
  - It would give many uses a significant speed boost, since we'd need to make less reads and writes
- Guiding users towards correct implementations
  - There are lots of warts in the spec, and we should let as few of them leak through our API as possible
  - This is text encoding, encryption underspecification, time handling, and whatever else I've missed
- Descriptive documentation!
  - This is something that we can address in the shorter term, no need to wait for a new API
  

```rust
mod zip {
    use std::{io, fs, path, time};
    
    pub struct Archive<S> {
        inner: S
    }
    impl<S: io::Read + io::Seek> Archive<S> {
        pub fn open(source: S) -> io::Result<Archive<S>>;
    }
    impl<S> Archive<S> {
        pub fn get(&mut self, path: impl Eq<PathBuf>) -> Option<File<S>>; 
        pub fn find(&mut self, predicate: impl FnMut(&File<S>) -> bool) -> Option<File<S>>;
    }
    impl<S: io::Write> Archive<S> {
        pub fn create(sink: S) -> Archive<S>;
        pub fn insert(&mut self, path: PathBuf) -> FileMut<S>;
    }
    pub enum CompressionMethod {
        Store,
        Shrink,
        Reduce1,
        Reduce2,
        Reduce3,
        Reduce4,
        Implode,
        Deflate,
        Deflate64,
        PkwareImplode,
        Bzip2,
        Lzma,
        IbmZosCmpsc,
        IbmTerse,
        Lz77,
        Zstd,
        Mp3,
        Xz,
        Jpeg,
        WavPack,
        PPMd,
        
        /// The data is stored in a proprietary PKWARE format
        Reserved,
    }
    /// Text from a ZIP file
    ///
    /// While ZIP files should contain either CP437 or UTF-8 encoded text, many
[...]

PR #215, opened 2020-12-23
Support append to an existing archive

Contextualist:
The implementation provides an alternative constructor, `new_append`, for `ZipWriter`, that takes a read-write buffer of existing archive and make it possible to add files to it.

`new_append` is mostly a copy of the `ZipArchive` constructor. It parses the central directory for the file list, and then set the cursor to the beginning of central directory in order to overwrite it. Users are expected to use the `ZipWriter` in the same way as if it is constructed from an empty buffer. When `finish` is called, the new central directory added will contains information for all prior files along with the appended files.

As the central directory parsing is borrowed from `ZipArchive`'s internal implementation, I need to make  `ZipArchive ::get_directory_counts` and `central_header_to_zip_file` public, which looks awkward to me. Is there a more graceful way? Also, suggestions on a better name for the constructor `new_append` are welcome.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -209,7 +209,7 @@ fn make_reader<'a>(
 impl<R: Read + io::Seek> ZipArchive<R> {
     /// Get the directory start offset and number of files. This is done in a
     /// separate function to ease the control flow design.
-    fn get_directory_counts(
+    pub fn get_directory_counts(
         reader: &mut R,
         footer: &spec::CentralDirectoryEnd,
         cde_start_pos: u64,
@@ -514,7 +514,8 @@ fn unsupported_zip_error<T>(detail: &'static str) -> ZipResult<T> {
     Err(ZipError::UnsupportedArchive(detail))
 }
 
-fn central_header_to_zip_file<R: Read + io::Seek>(
+/// Parse a central directory entry to collect the information for the file.
+pub fn central_header_to_zip_file<R: Read + io::Seek>(
     reader: &mut R,
     archive_offset: u64,
 ) -> ZipResult<ZipFileData> {
--- a/src/read.rs
+++ b/src/read.rs
@@ -209,7 +209,7 @@ fn make_reader<'a>(
 impl<R: Read + io::Seek> ZipArchive<R> {
     /// Get the directory start offset and number of files. This is done in a
     /// separate function to ease the control flow design.
-    pub fn get_directory_counts(
+    pub(crate) fn get_directory_counts(
         reader: &mut R,
         footer: &spec::CentralDirectoryEnd,
         cde_start_pos: u64,
@@ -515,7 +515,7 @@ fn unsupported_zip_error<T>(detail: &'static str) -> ZipResult<T> {
 }
 
 /// Parse a central directory entry to collect the information for the file.
-pub fn central_header_to_zip_file<R: Read + io::Seek>(
+pub(crate) fn central_header_to_zip_file<R: Read + io::Seek>(
     reader: &mut R,
     archive_offset: u64,
 ) -> ZipResult<ZipFileData> {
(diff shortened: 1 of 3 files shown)


copyandexecute:
Great work! I would really looking forward to it accepting this SPECIFIC pullrequest since I need it for my work

copyandexecute:
bump

F0Xde:
I would suggest just naming the function `append`, I don't think the additional `new` prefix is really required for a good understanding of what it does, but the current name is also totally fine.

copyandexecute:
bump

src/read.rs
```diff
@@ -209,7 +209,7 @@ fn make_reader<'a>(
 impl<R: Read + io::Seek> ZipArchive<R> {
     /// Get the directory start offset and number of files. This is done in a
     /// separate function to ease the control flow design.
-    fn get_directory_counts(
+    pub fn get_directory_counts(
```
Plecra, line 212: `pub(crate)` will do the trick!

src/write.rs
```diff
@@ -194,6 +194,46 @@ impl ZipWriterStats {
     }
 }
 
+impl<A: Read + Write + io::Seek> ZipWriter<A> {
+    /// Initializes the archive from an existing ZIP archive, making it ready for append.
+    pub fn new_append(mut readwriter: A) -> ZipResult<ZipWriter<A>> {
```
Plecra, line 199: I'd like to simply call this `open`, but I think we can bikeshed it for a bit before release

src/write.rs
```diff
@@ -194,6 +194,46 @@ impl ZipWriterStats {
...
+        let mut files = Vec::new();
+
+        if let Err(_) = readwriter.seek(io::SeekFrom::Start(directory_start)) {
+            return Err(ZipError::InvalidArchive(
+                "Could not seek to start of central directory",
+            ));
+        }
+
+        for _ in 0..number_of_files {
+            let file = central_header_to_zip_file(&mut readwriter, archive_offset)?;
```
Plecra, line 220: Please use an iterator here. Roughly
```rust
let files = (..number_of_files)
    .map(|_| central_header_to_zip_file(&mut readwriter, archiveoffset))
    .collect::<Result<Vec<_>, _>>()?;
```

Plecra (changes requested):
I like the look of this! I'll need to test a couple things soon, but there shouldn't be any problem :)

src/write.rs
Contextualist: Will keep this for an overall API change, since currently both the `ZipArchive` constructor and `ZipWriter<Write + Seek>` constructor have the name `new`.

Contextualist:
Thanks for your suggestions @Plecra! I learn something useful.

src/read.rs
Contextualist: That's nice. BTW, I saw in another PR that you mention about the current situation of having various edge-case API sprinkled around, and I really appreciate your works in #156 in your limited time. I'll see if I can help.

src/write.rs
```diff
@@ -194,6 +194,43 @@ impl ZipWriterStats {
...
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let _ = readwriter.seek(io::SeekFrom::Start(directory_start)); // seek directory_start to overwrite it
+
+        Ok(ZipWriter {
+            inner: GenericZipWriter::Storer(readwriter),
+            files,
+            stats: Default::default(),
+            writing_to_file: false,
+            comment: String::new(),
```
Contextualist, line 228: I should probably pick up the comment from `footer.zip_file_comment`, like what `ZipArchive::new` is doing. However, I found that `ZipWriter.comment` is a `String`, unlike `ZipArchive.comment` which is a `Vec<u8>`, so we cannot directly store the bytes from `footer.zip_file_comment`. Should I assume those bytes to be UTF-8 text and handle the conversion error?

Plecra:
Thanks for the new API @Contextualist ! This looks great ~

PR #220, opened 2021-01-29
Update file_info.rs

pcrumley:
Update docs to get rid of `enclosed_name` function which no longer exists.


pcrumley:
sorry i should not try to code rust in github editor. 😅

Issue #25, opened 2017-01-25
Can ZipFile implement Seek?


icefoxen:
I feel like this might have been asked before; if so, I apologize.

Is it possible for the `ZipFile` type to implement the `Seek` trait?  I'm honestly not sure, but it would be useful to have the option, even if it involves iterating through the whole file decompressing it on the fly.

Thanks.

icefoxen:
Related to issue #14 :/

mvdnes:
This is not actually related to #14.
For compressed files, it would not be practical to implement seek. Zlib only allows a stream to move forward. So unless you decompress the entire file in memory, you cannot go backwards.
For small files this would be OK, but this library should also take files into account that could be very large.

If you want to implement this yourself, you can read the file into a Vec, and then wrap in in a [Cursor].

[Cursor]: https://doc.rust-lang.org/stable/std/io/struct.Cursor.html

zumoshi:
@mvdnes What about seeking only for files using `Store` (i.e. no compression)
My use case is bundling some metadata with an encrypted file (which won't compress well anyway), but I need seeking inside that file. Would you recommend another file format or is it possible to get a seekable slice of the zip file from this library?

PR #221, opened 2021-01-30
impl Seek for ZipFile (uncompressed/unencrypted files only)

zumoshi:
Proof of concept only. will need tests/refactoring before merging.
This is just mostly asking if you consider this feature useful to have.

example use:
```rust
fn main() {
    // extracting part of test.zip/test into B:/test.txt
    let fname = std::path::Path::new("./test.zip");
    let mut dst = File::create("B:/test.txt").unwrap();
    // from examples/extract.rs
    let zipfile = std::fs::File::open(&fname).unwrap();
    let mut archive = zip::ZipArchive::new(zipfile).unwrap();

    // will error if encrypted/compressed
    let mut file = match archive.by_name_seekable("test") {
        Ok(file) => file,
        Err(..) => {
            panic!("File test not found or cannot seek");
        }
    };
    // chunked copy (not needed in this case but was needed for my use case of >1GB files)
    const chunk_size: usize = 1024;
    let mut buffer: [u8; chunk_size] = [0; chunk_size];

    // seeking (will error if not seekable stream)
    file.seek(SeekFrom::Start(20000u64)).unwrap();
    let mut read = 20000;
    let mut wrote = 20000;

    while read < file.size() as usize {
        read += file.read(&mut buffer).unwrap();
        wrote += dst.write(&buffer[0..(read - wrote)]).unwrap();
    }
}
```

My use case for wanting this is to bundle some metadata (a JSON file) with my encrypted binary blob which has information like where to get the key for decryption. after which I will need to read the encrypted file. due to its size of >1GB extracting would have too much overhead, and other solutions like partial encryption or changing the file format itself increases complexity.
My current solution is to zip a JSON metadata file and the binary file with store mode (it wouldn't compress well anyway since it's encrypted) and the only missing piece was a Seek+Read reference into it without needing to extract.

Solves #25.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -8,7 +8,7 @@ use crate::zipcrypto::ZipCryptoReader;
 use crate::zipcrypto::ZipCryptoReaderValid;
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::io::{self, prelude::*};
+use std::io::{self, prelude::*, SeekFrom};
 use std::path::{Component, Path};
 
 use crate::cp437::FromCp437;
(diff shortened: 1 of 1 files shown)


Plecra:
I am going to be updating the [`ZipArchive::by_index_raw`](https://docs.rs/zip/0.5.9/zip/read/struct.ZipArchive.html#method.by_index_raw) API to return a special `RawReader` type, which should certainly have a `Seek` implementation!

I don't want to implement the feature like this, but could you create a new issue for seeking through raw files?

PR #231, opened 2021-04-10
Remove `io::Read + io::Seek` bound to `ZipArchive`

a1phyr:
This is more in line with what is usually done and easier to use for downstream users.


Plecra:
Cheers!

PR #239, opened 2021-05-17
Add method for writing files with raw byte filenames.

cessen:
This allows handling e.g. properly round-tripping zip files with contents that have non-utf8 filepaths.

I ran into this with a tool I'm writing, that specifically needs to work with zip files that contain alternative filename encodings, both reading and writing.  It's admittedly a little niche, but it does come up occasionally, and seems easy enough to accommodate.

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -1067,11 +1067,9 @@ fn write_local_file_header<T: Write>(writer: &mut T, file: &ZipFileData) -> ZipR
     writer.write_u32::<LittleEndian>(spec::LOCAL_FILE_HEADER_SIGNATURE)?;
     // version needed to extract
     writer.write_u16::<LittleEndian>(file.version_needed())?;
-    // general purpose bit flag
-    let flag = if !std::str::from_utf8(&file.file_name_raw)
-        .unwrap_or("あ")
-        .is_ascii()
-    {
+    // "General Purpose" bit flag.  When set, indicates that the
+    // name is utf8 compatible (either ascii or utf8-encoded unicode).
+    let flag = if !std::str::from_utf8(&file.file_name_raw).is_ok() {
         1u16 << 11
     } else {
         0
--- a/src/write.rs
+++ b/src/write.rs
@@ -429,7 +429,8 @@ impl<W: Write + io::Seek> ZipWriter<W> {
         }
         *options.permissions.as_mut().unwrap() |= 0o100000;
         self.start_entry(name.into().as_bytes(), options, None)?;
-        self.inner.switch_to(options.compression_method, options.compression_level)?;
+        self.inner
+            .switch_to(options.compression_method, options.compression_level)?;
         self.writing_to_file = true;
         Ok(())
     }
@@ -1145,11 +1146,9 @@ fn write_central_directory_header<T: Write>(writer: &mut T, file: &ZipFileData) 
     writer.write_u16::<LittleEndian>(version_made_by)?;
     // version needed to extract
     writer.write_u16::<LittleEndian>(file.version_needed())?;
-    // general puprose bit flag
-    let flag = if !std::str::from_utf8(&file.file_name_raw)
-        .unwrap_or("あ")
-        .is_ascii()
-    {
+    // "General Purpose" bit flag.  When set, indicates that the
+    // name is utf8 compatible (either ascii or utf8-encoded unicode).
+    let flag = if !std::str::from_utf8(&file.file_name_raw).is_ok() {
         1u16 << 11
     } else {
         0
--- a/src/write.rs
+++ b/src/write.rs
@@ -4,7 +4,7 @@ use crate::compression::CompressionMethod;
 use crate::read::{central_header_to_zip_file, ZipArchive, ZipFile};
 use crate::result::{ZipError, ZipResult};
 use crate::spec;
-use crate::types::{AtomicU64, DateTime, System, ZipFileData, DEFAULT_VERSION};
+use crate::types::{AtomicU64, DateTime, MaybeUtf8, System, ZipFileData, DEFAULT_VERSION};
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use crc32fast::Hasher;
 use std::default::Default;
@@ -27,41 +27,6 @@ use time::OffsetDateTime;
 
 #[cfg(feature = "zstd")]
 use zstd::stream::write::Encoder as ZstdEncoder;
-
-/// Stores either a UTF8 string or raw bytes.
-///
-/// The reason for using this rather than just always using raw bytes
-/// is to communicate intent: the `Raw` variant means that the data
-/// is explicitly not meant to be valid utf8, even if it can
-/// coincidentally be interpreted that way.
-#[derive(Debug, Copy, Clone, Eq, PartialEq)]
-enum MaybeUtf8<'a> {
-    Utf8(&'a str),
-    Raw(&'a [u8]),
-}
-
-impl<'a> MaybeUtf8<'a> {
-    // fn as_str(&self) -> Option<&str> {
-    //     match *self {
-    //         MaybeUtf8::Utf8(string) => Some(string),
-    //         MaybeUtf8::Raw(_) => None,
-    //     }
-    // }
-
-    fn as_bytes(&self) -> &[u8] {
-        match *self {
-            MaybeUtf8::Utf8(string) => string.as_bytes(),
-            MaybeUtf8::Raw(bytes) => bytes,
-        }
-    }
-
-    fn to_string_lossy(&self) -> std::borrow::Cow<str> {
-        match *self {
-            MaybeUtf8::Utf8(string) => string.into(),
-            MaybeUtf8::Raw(bytes) => String::from_utf8_lossy(bytes),
-        }
-    }
-}
 
 enum GenericZipWriter<W: Write + io::Seek> {
     Closed,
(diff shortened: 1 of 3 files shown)


BenjaminRi:
Naive question: Why not use `std::path::Path` ( https://doc.rust-lang.org/std/path/struct.Path.html ) which is specifically designed for this purpose?

cessen:
That would make a lot of sense, yeah.  I made it take bytes because that's what `zip::read::ZipFile::name_raw()` returns, and it seems like these two methods should be symmetric.

I'd be happy to change both this new method and `name_raw()` to take/return `Path` instead.  How does that sound?

BenjaminRi:
I've been thinking about it, and `std::path::Path` brings in a needless dependency on `std::ffi::OsStr` and doesn't solve the problem at hand. So, scrap that idea.

About the issue of encoding: The latest ZIP standard ( https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT ) says the following about file name encoding:

```

D.1 The ZIP format has historically supported only the original IBM PC character 
encoding set, commonly referred to as IBM Code Page 437.  This limits storing 
file name characters to only those within the original MS-DOS range of values 
and does not properly support file names in other character encodings, or 
languages. To address this limitation, this specification will support the 
following change. 

D.2 If general purpose bit 11 is unset, the file name and comment SHOULD conform 
to the original ZIP character encoding.  If general purpose bit 11 is set, the 
filename and comment MUST support The Unicode Standard, Version 4.1.0 or 
greater using the character encoding form defined by the UTF-8 storage 
specification.  The Unicode Standard is published by the The Unicode
Consortium (www.unicode.org).  UTF-8 encoded data stored within ZIP files 
is expected to not include a byte order mark (BOM). 
```

This is problematic for a number of reasons. First of all, the language used here: `If general purpose bit 11 is unset, the file name and comment SHOULD conform to the original ZIP character encoding` means that all bets are off if this bit is not set. Anyone can use anything because `SHOULD` is not mandatory. So, by default, we should never emit files like that.

`If general purpose bit 11 is set, the filename and comment MUST support The Unicode Standard, Version 4.1.0 or greater using the character encoding form defined by the UTF-8 storage specification` by default, we always want to emit ZIP files like this. It is the only way for the file name not to be ill-specified.

Ideally, our API would let the user decide between the clean UTF-8 solution (with general purpose bit 11 set) or the ill-specified solution (without general purpose bit 11) and if the user chooses the ill-specified solution, he should at least be nudged into using CP437. Because the characters of CP437 are a subset of Unicode, I believe that the best approach would be to keep the file name a Rust String, but let the user choose whether he writes UTF-8 (also set the general purpose bit 11) or whether he writes CP437 (without general purpose bit 11). Because CP437 maps the entire byte to characters, (unlike e.g. ASCII that only maps 7 bit) this also allows for arbitrary binary filenames, but this solution would force the user to go through this encoding, which I believe to be a win because it makes it harder to violate the `SHOULD` mentioned in the standard above.

Currently, our logic auto-detects ASCII strings and does not set the general purpose bit 11 if it's just ASCII:

```rust
    let flag = if !file.file_name.is_ascii() {
        1u16 << 11
    } else {
        0
    };
```

So it's not controllable by the API user. I wonder what's the point of this because I would always set the general purpose bit 11 if we have an UTF-8 string (even if it's just the ASCII subset), so we clearly signal what encoding we use. I wonder if the above auto-detection was deliberately designed for compatibility reasons. In that case, I wonder what repercussions systematically setting general purpose bit 11 would have.

Also, I noticed something in your PR that definitely cannot stay: `.unwrap_or("あ")`. Must have slipped in somehow. This has to be done differently. Also, it can lead to an erroneous case of general purpose bit 11 being set when in fact the user intended to use CP437.

I agree that the problem you are having (file name encoding, round-tripping file names) is an issue we need to solve, but I think this PR needs work and careful consideration so we don't start violating the ZIP spec.

BenjaminRi:
Addendum: Looks like 7-zip also unsets general purpose bit 11 for plain ASCII strings in some cases:

https://github.com/kornelski/7z/blob/a6e2c7e401a3e5976e8522de518a169b0d8a7fac/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp#L256

So it seems like this behaviour is not unique to `zip-rs`. If this is somehow the recommended way to do it, I would still keep the file name a Rust `String` and present the user with the choice (with a new enum) between UTF-8 or CP437 writing, if he chooses UTF-8, the general purpose bit 11 will be set unless the string is only ASCII. And for CP437 it will never be set.

cessen:
> Also, I noticed something in your PR that definitely cannot stay: `.unwrap_or("あ")`.

Oops!  Yeah, that was something I tossed in while I was just trying to get things working for my project, and I totally forgot to revisit it before the PR.

> The latest ZIP standard ( https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT ) says the following about file name encoding:
>
> [...]means that all bets are off if this bit is not set.

So in other words, if we just never the set the bit, we're technically conforming no matter what's in the filename?  But if the bit *is* set, then it must be utf8?  In other words, the bit being set means "guaranteed utf8" and it not being set means "can't know anything for sure".

From a historical perspective that makes sense.  Text encodings are a deep and crazy topic, and even to this day many people in Japan, for example, use Shift-JIS.  And BIG5 in China.  Etc.  So I can imagine there are a lot of historical zip files with arbitrary filename encodings out there.  So if the standard said "must" instead of "should", it would be breaking a lot of data in practice.  Setting the bit is kind of a way to make a clean break for new files.

> I wonder what's the point of this because I would always set the general purpose bit 11 if we have an UTF-8 string (even if it's just the ASCII subset), so we clearly signal what encoding we use.

Yeah, that makes sense to me.  At least based on the spec, it seems like it would be both easier and more correct to just validate for utf8 when writing, and if the filename is valid utf8 set the bit, and if it's not then unset it.

> So it's not controllable by the API user.

I think that makes sense.  I mean, maybe add control for it if a use-case for manually setting it comes up.  But that's certainly not a use-case I have.

cessen:
Seems I'm not alone in running into these kinds of issues:
https://github.com/xTibor/fix-japanese-zip

(Although my need for round-tripping might be more niche.)

zamazan4ik:
@cessen maybe you have some updates on the PR?

cessen:
Ah!  Sorry, I haven't actually done any more work on this.  My personal use case was a project that I'm no longer working on, so I never came back to this.

But I'm happy to get it into shape for merging if the feature is still wanted by others.  I should have some time in the next couple of weeks or so to work on it.

src/write.rs
```diff
@@ -984,7 +998,10 @@ fn write_central_directory_header<T: Write>(writer: &mut T, file: &ZipFileData)
     // version needed to extract
     writer.write_u16::<LittleEndian>(file.version_needed())?;
     // general puprose bit flag
-    let flag = if !file.file_name.is_ascii() {
+    let flag = if !std::str::from_utf8(&file.file_name_raw)
```
Plecra, line 1001: ''

src/write.rs
```diff
@@ -897,7 +908,10 @@ fn write_local_file_header<T: Write>(writer: &mut T, file: &ZipFileData) -> ZipR
     // version needed to extract
     writer.write_u16::<LittleEndian>(file.version_needed())?;
     // general purpose bit flag
-    let flag = if !file.file_name.is_ascii() {
+    let flag = if !std::str::from_utf8(&file.file_name_raw)
+        .unwrap_or("あ")
```
Plecra, line 912: Hah, this took a moment to understand. A fun solution, but I'd prefer `file.file_name_raw.is_ascii()` ;)

src/write.rs
Plecra, line 1001: ''

Plecra (approved):
This is lovely. I appreciate the small changes to the API, and the implementation is tidy. Perhaps we should add a `start_file_raw_name_with_extra_data` (ew 😆) later.

cessen:
I still haven't got around to this (sorry!).  Just dropping a line to say that it's still on my radar, and I haven't forgotten about it.  I just ended up being busier than I expected with other things.  I'll try to get to this soon-ish.

Plecra:
Yeah no worries! Don't rush to it, there are still some problems with assumptions about the encoding used. It's tricky to fix...

cessen:
I've removed the stupid `あ` hack, and now set the general purpose bit whenever the filename is valid utf8 (as per the spec).

> there are still some problems with assumptions about the encoding used

Yeah, I feel like there's a bit of tension here between providing maximum control, and presenting a "just works" API to the user.  

One possibility is to just set the general purpose bit based on which method the user calls.  Using `start_file()` would always set the general purpose bit, since it's guaranteed to be utf8, and using `start_file_raw_name()` would always *not* set the bit, with the assumption that users would only call that when using non-utf8 encodings.  Since most users will just use `start_file()` anyway, that keeps the happy path easy and simple.  And then for the niche cases when alternative encodings are desired `start_file_raw_name()` can be used.

What do you guys think?

Plecra:
Yup, I'd prefer that latter solution. In practice, it looks like no tools actually use the `utf8` bit anyway, so I expect it'll often be fine not to set it

cessen:
Okay, I think I've got this in a good state now, modulo issues found by review.  It required a little bit of reworking some existing code, but I don't think it's too bad.  And it will make it easier to add additional methods in the future (e.g. if we want to support raw comments, etc.).

zamazan4ik:
@Plecra we need your review here)

Plecra (approved):
Urgh, still not totally happy with the inconsistencies with the ZIP format, but that's how it goes with so many nonconforming implementations out there :)

Our solution here will have to do for now. I'd like to start with the `unsafe` module treatment for this and see if that'll cover users' needs. Thankyou for the PR :) 

(checking the merge conflicts, it's an easy fix. bring the capacity check into this new code)

Plecra:
I'll rework this one to only use on `file_names` map. Sorry I didn't make use of this one cessen :)

cessen:
Yeah, no worries!  As long as the functionality gets in one way or another, I'm happy. :-)

PR #244, opened 2021-07-05
chore: bump rand to 0.8

striezel:
The Minimum Supported Rust Version of rand 0.8 is rustc 1.36, which is exactly the MSRV of zip. Therefore, this should be safe to bump.


zamazan4ik:
@striezel Thanks!

striezel:
It's good to see that things are moving forward again with this crate. :)

PR #250, opened 2021-09-06
Remove dependency on thiserror

linkmauve:
On my computer, this halves the total build time from 11.7s to 5.4s, as well as the total size of the artifacts in target.

This derive macro is nice, but it doesn’t justify the increase in compilation time for dependent crates imo.

This effectively reverts #135, although using the up to date APIs from `std::error::Error`.


src/result.rs
```diff
@@ -1,37 +1,66 @@
...
+impl From<io::Error> for ZipError {
+    fn from(err: io::Error) -> ZipError {
+        ZipError::Io(err)
+    }
+}
+
+impl fmt::Display for ZipError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            ZipError::Io(err) => write!(fmt, "{}", err),
```
a1phyr, line 47: The `io::Error` is returned by `source`, so according to [the guidelines](https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html#guidelines-for-implementing-displayfmt-and-errorsource) it probably shouldn't be printed here.

You could change the message to something like `"IO error"`, I think.

zamazan4ik:
Honestly, I am not a big fan of such changes since it doesn't affect significantly the UX for end-users. For local development, a library is usually built only once. Maybe the most affected area here is CI but for CI some kind of cache should be used.

If we use some existing libraries instead of reimplementing the same functionality, again and again, it improves code reusability across the whole ecosystem. Also, if some changes will be made to the upstream `thiserror` crate, we will fetch them automatically with a usual `cargo upgrade` during the maintenance.

Plecra (approved):

PR #251, opened 2021-09-06
Use getrandom instead of rand for benches

linkmauve:
The current code didn’t build, and this one includes fewer dependencies than the full rand set of crates. This isn’t any slower from a quick benchmark on a Kaby Lake CPU, although I get a quite high variance in both cases.

Together with #250, this brings down the number of crates to build from 44 to 28.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ thiserror = "1.0"
 
 [dev-dependencies]
 bencher = "0.1"
-rand = "0.7"
+getrandom = "0.2"
 walkdir = "2"
 
 [features]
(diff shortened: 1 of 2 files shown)


zamazan4ik:
@linkmauve Can you please resolve the conflicts?

zamazan4ik:
I have checked the documentation for both `getrandom` and `rand` crates. According to my findings, in this case, it's acceptable to use `getrandom` since doesn't introduce any drawbacks and only lowers build time.

zamazan4ik (approved):

linkmauve:
Thanks!

PR #258, opened 2022-01-22
feat: bump MSRV

zamazan4ik:
- update MSRV to 1.54 according to the policy. Seems like it needed for
  building dependencies properly
- update CI

Tested:
- No

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        rust: [stable, 1.36.0]
+        rust: [stable, 1.54.0]
 
     steps:
     - uses: actions/checkout@master
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ All of these are enabled by default.
 MSRV
 ----
 
-Our current Minimum Supported Rust Version is **1.36.0**. When adding features,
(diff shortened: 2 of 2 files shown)


zamazan4ik:
It's required for fixing the current build in master. Also, it will help us easier upgrade dependencies in the near future.

kirillbobyrev:
I would also suggest possibly bumping MSRV to 1.58.1 which was just released to mitigate [CVE-2022-21658](https://www.cve.org/CVERecord?id=CVE-2022-21658) (more [info](https://blog.rust-lang.org/2022/01/20/Rust-1.58.1.html)).

zamazan4ik:
I guess it should be "up to user" to choose the Rust version for their programs. For us, there is no reason to force only the latest Rust compiler version.

kirillbobyrev:
I kind of agree in that zip is a library, so if my mental model is correct, the users will recompile it with their MSRV either way which rules this CVE out. However, 1.58.1 is likely to be quite popular because of that CVE, so it might just be a bit more convenient :) Up to you, this is unlikely to be super important.

djc:
The policy seems a little unclear to me. The latest 4 minor updates right now means 1.54.0, but that's not quite 6 months yet -- 6 months ago 1.54 had not been released.

zamazan4ik:
I think "6 months" here you can interpret just as an approximation. So the policy actually is "4 latest Rust releases" and you should rely on it, not on provided time span.

PR #259, opened 2022-01-22
fix: add header offset overflow check

zamazan4ik:
- during the header offset calculation, perform overflow check

Tested:
- Local tests


zamazan4ik:
A little later I want to add additional test case for the issue.

tests/issue_234.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[test]
+fn invalid_date() {
+    let reader = std::io::Cursor::new(&BUF);
+    let archive = zip::ZipArchive::new(reader);
+    match archive {
+        Err(ZipError::InvalidArchive(_)) => {}
+        _ => {
+            panic!(
+                "Unexpected result: expected {:?}",
```
5225225, line 31: Might be worth printing what the result you *did* get was?

If you're debugging this, you'd end up changing the code to include that anyways.

zamazan4ik: Good point. I will fix it in the next commit. Thanks!

zamazan4ik: Fixed

PR #262, opened 2022-01-23
fix: fix Clippy warnings

zamazan4ik:
- fix a bunch of Clippy warnings
- fix some usages of assert! (change to assert_ne)

Tested:
- Local unit-tests run

### DIFF
--- a/src/compression.rs
+++ b/src/compression.rs
@@ -125,7 +125,7 @@ mod test {
 
     #[test]
     fn from_eq_to() {
-        for v in 0..(::std::u16::MAX as u32 + 1) {
+        for v in 0..(u16::MAX as u32 + 1) {
             #[allow(deprecated)]
             let from = CompressionMethod::from_u16(v as u16);
             #[allow(deprecated)]
@@ -135,17 +135,17 @@ mod test {
     }
 
     fn methods() -> Vec<CompressionMethod> {
-        let mut methods = Vec::new();
-        methods.push(CompressionMethod::Stored);
-        #[cfg(any(
-            feature = "deflate",
-            feature = "deflate-miniz",
-            feature = "deflate-zlib"
-        ))]
-        methods.push(CompressionMethod::Deflated);
-        #[cfg(feature = "bzip2")]
-        methods.push(CompressionMethod::Bzip2);
-        methods
+        vec![
+            CompressionMethod::Stored,
+            #[cfg(any(
+                feature = "deflate",
+                feature = "deflate-miniz",
+                feature = "deflate-zlib"
+            ))]
+            CompressionMethod::Deflated,
+            #[cfg(feature = "bzip2")]
+            CompressionMethod::Bzip2,
+        ]
     }
 
     #[test]
--- a/src/read.rs
+++ b/src/read.rs
@@ -189,11 +189,11 @@ fn make_crypto_reader<'a>(
     Ok(Ok(reader))
 }
 
-fn make_reader<'a>(
+fn make_reader(
     compression_method: CompressionMethod,
     crc32: u32,
-    reader: CryptoReader<'a>,
-) -> ZipFileReader<'a> {
+    reader: CryptoReader,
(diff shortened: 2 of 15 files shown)


zamazan4ik:
I didn't fix yet all Clippy warnings since for some of the, that's a little questionable, how to fix them properly.

src/compression.rs
```diff
@@ -134,6 +134,7 @@ mod test {
         }
     }
 
+    #[allow(clippy::vec_init_then_push)]
     fn methods() -> Vec<CompressionMethod> {
```
Plecra, line 138: I think clippy wants code like this, which is an alright improvement.
```rust
vec![
    CompressionMethod::Stored
    #[cfg(any(
        feature = "deflate",
        feature = "deflate-miniz",
        feature = "deflate-zlib"
    ))]
    CompressionMethod::Deflated,
    #[cfg(feature = "bzip2")]
    CompressionMethod::Bzip2,
]
```

zamazan4ik: Just was not sure is it nice enough :) Will fix it.

zamazan4ik: As I thought:
```
rror: no rules expected the token `#`
   --> src/compression.rs:140:13
    |
140 |             #[cfg(any(
    |             ^ no rules expected this token in macro call

error: could not compile `zip` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed

```

src/read.rs
```diff
@@ -457,14 +457,14 @@ impl<R: Read + io::Seek> ZipArchive<R> {
     }
 
     /// Get a contained file by index
-    pub fn by_index<'a>(&'a mut self, file_number: usize) -> ZipResult<ZipFile<'a>> {
+    pub fn by_index(&mut self, file_number: usize) -> ZipResult<ZipFile> {
```
Plecra, line 460: Can you add the elided lifetime for `ZipFile<'_>`? The borrow is an important part of the signature and iirc, it affects the docs.

src/types.rs
```diff
@@ -88,6 +88,7 @@ impl DateTime {
     /// * hour: [0, 23]
     /// * minute: [0, 59]
     /// * second: [0, 60]
+    #[allow(clippy::result_unit_err)]
```
Plecra, line 91: A better error type will be nice when we can fix the api

src/types.rs
```diff
@@ -329,15 +327,15 @@ mod test {
         use super::DateTime;
         let dt = DateTime::default();
         assert_eq!(dt.timepart(), 0);
-        assert_eq!(dt.datepart(), 0b0000000_0001_00001);
+        assert_eq!(dt.datepart(), 0b0000_0000_0010_0001);
```
Plecra, line 330: The formatting here is useful: `datepart` has three bitfields with different lengths, so using runs of 4 is confusing. (That's the first of January. Not so obvious with the new formatting ;))

Plecra (changes requested):
Marvelous! Clippy's recommendations were mostly great. Even caught some nasty bugs in there, which is nice to see. Only a few style 'regressions' you might want to reconsider.

src/types.rs
zamazan4ik: Sure. That's why I've commented it out for now :) Later, when we will rework an API, we definitely will fix it (at least I hope)

src/compression.rs
zamazan4ik: Nvm, fixed - my mistake

src/read.rs
zamazan4ik: Done

src/compression.rs
zamazan4ik: Fixed

src/types.rs
zamazan4ik: Fixed

Plecra (approved):

Plecra (approved):

Issue #252, opened 2021-09-29
Permissions are not copied in raw_copy_file_rename


dzamlo:
When using the ` raw_copy_file_rename` method (and the ` raw_copy_file` method that call it), the permission are not copied.

The issue is with this line:
https://github.com/zip-rs/zip/blob/7edf2489d5cff8b80f02ee6fc5febf3efd0a9442/src/write.rs#L610

When calling the setters on `FileOptions`, the original one is not modified  because the `struct` is `Copy` and thus the method is called with an implicit copy and not a reference to the original one.

zamazan4ik:
@dzamlo Thanks for the issue? Maybe you already have the proper fix for the library? Any PR is welcome!

zamazan4ik:
Fixed in https://github.com/zip-rs/zip/pull/264

PR #264, opened 2022-01-23
fix: permissions are not copied

zamazan4ik:
- fix a small bug when permissions where not copied

Tested:
- No

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -133,6 +133,7 @@ impl FileOptions {
     ///
     /// The default is `CompressionMethod::Deflated`. If the deflate compression feature is
     /// disabled, `CompressionMethod::Stored` becomes the default.
+    #[must_use]
     pub fn compression_method(mut self, method: CompressionMethod) -> FileOptions {
         self.compression_method = method;
         self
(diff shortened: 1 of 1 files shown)


zamazan4ik:
Fixes https://github.com/zip-rs/zip/issues/252

Plecra (approved):

Plecra:
I wonder if we could make this easier to catch in future reviews, or just need to remember. Would `#[must_use]` on `FileOptions` do it?

zamazan4ik:
Maybe this thread will be helpful for us: https://github.com/rust-lang/rust/issues/48926

zamazan4ik:
Maybe another option is divide this struct into `FileOptions` and `FileOptionsBuilder` or smth like that

Plecra:
Ah, you're right, that issue does help. We ought to just add `must_use` to each of the methods on `FileOptions`. Then, it's alright for other functions to return an ignored `FileOptions`. Can you add that in this PR please?

zamazan4ik:
Sure, will do it

zamazan4ik:
Done for all methods, except `default`.

PR #267, opened 2022-01-24
fix: remove extra CompressionMethod variant

### LINKED ISSUE #266: Accidental API break in ZSTD PR
https://github.com/zip-rs/zip/blob/afc84af4aafa2af8c4b1d10693d366b6c640002f/src/compression.rs#L27-L29

This enum isn't `non_exhaustive` so this change is breaking. We need to remove it and update the code to use the existing `CompressionMethod::ZSTD` (and `CompressionMethod::ZSTD_DEPRECATED`)

Plecra:
this is required to maintain backwards compatibility with 0.5

Closes #266 

I doubt these changes would introduce any problems themselves. @zamazan4ik, could you just check it's still well behaved with your zlib files? We'll add better tests before the release

### DIFF
--- a/examples/write_dir.rs
+++ b/examples/write_dir.rs
@@ -33,7 +33,7 @@ const METHOD_BZIP2: Option<zip::CompressionMethod> = Some(zip::CompressionMethod
 const METHOD_BZIP2: Option<zip::CompressionMethod> = None;
 
 #[cfg(feature = "zstd")]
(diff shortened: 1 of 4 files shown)


Plecra:
Hm. That test failure makes sense. Need to decide how to handle `ZSTD_DEPRECATED` zip archives that we find.

Plecra:
The new changes have diverged from the 0.5 API enough that a breaking version bump is needed, making this work unnecessary

Issue #243, opened 2021-07-02
Question: about the decompress speed on Arm


chirping78:
A very simple program very like the example:
```
                for i in 0..archive.len() {
                    let mut file = archive.by_index(i).unwrap();
                    let outpath = file.enclosed_name().unwrap();
                    let mut outfile = std::fs::File::create(parent.join(outpath)).unwrap();
                    debug!("extract: {} ==> {}", outpath.display(), parent.join(outpath).display());
                    std::io::copy(&mut file, &mut outfile).unwrap();
                }

```
I use this program to decompress a file with about 400MB size, which contains 2~3 200MB files.
The speed is normal on X86 pc, but is very very slow on an Arm board.
On this Arm board, the busybox unzip takes about 20s to decompress, but the rust version has run about 20 minitues, still not finished...

flxo:
@chirping78 can you post the output of strace when running your test? e.g `strace -f -s 256 <your binary>.

I've seen a lot 1 byte reads done by zip-rs when reading a file from the archive - even when the reader is a `BufRead`.

chirping78:
I can reproduce the issue with the original example/extract:
```
 cargo b --target=armv7-unknown-linux-gnueabihf --example extract
```
when using the `extract` to decompress the file, the output speed is about 200kB / sec.
Since I'm using armv7 as the target, I think you can also reproduce it on raspberrypi, beagle, etc...

chirping78:
@flex Ok, will try your suggestion after short while...

> @chirping78 can you post the output of strace when running your test? e.g `strace -f -s 256 .
> 
> I've seen a lot 1 byte reads done by zip-rs when reading a file from the archive - even when the reader is a `BufRead`.

chirping78:
I have captured the strace log, it's too long to paste here, but the pattern is repeated. Paste some part here:
```
read(3, "\375\277\373\357\227 ....)=32768
write(4, "\2271\357\316?\307\   ...)=8192
write(4, "\3713\365<\224\207B  ...)=8192
write(4, "]\252\343\10\326\17
write(4, "\23\320N\235l\322:\
write(4, "\345\252\363j\340\2
read(3, ";}{\255\274)d\264\37
write(4, "\2019\347\6\37\2469
write(4, "\303\210M_\266.\217
write(4, "gVhM\235\320\357\21
write(4, "\370n\372\331\244\3
write(4, "f{\356UB\316b\26\33
read(3, "Q|\300\274~\360PB'\2
[...]

PR #270, opened 2022-01-24
feat: migrate to Criterion

zamazan4ik:
- get rid of bencher since it's outdated and is not updated anymore
- add more benches from https://github.com/kauhat/zip-rs/pull/1/
- add strum for enum enumeration

Tested:
- Local bench runs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,10 +16,12 @@ time = { version = "0.3", features = ["formatting", "macros" ], optional = true 
 byteorder = "1.3"
 bzip2 = { version = "0.4", optional = true }
 crc32fast = "1.1.1"
+strum = "0.23.0"
+strum_macros = "0.23.0"
 zstd = { version = "0.10", optional = true }
 
 [dev-dependencies]
-bencher = "0.1"
+criterion = "0.3.5"
 getrandom = "0.2"
 walkdir = "2"
 
--- a/benches/read_entry.rs
+++ b/benches/read_entry.rs
@@ -1,18 +1,21 @@
-use bencher::{benchmark_group, benchmark_main};
+use criterion::{criterion_group, criterion_main};
+use criterion::{BenchmarkId, Criterion};
 
 use std::io::{Cursor, Read, Write};
 
-use bencher::Bencher;
 use getrandom::getrandom;
-use zip::{ZipArchive, ZipWriter};
+use strum::IntoEnumIterator;
+use zip::{CompressionMethod, ZipArchive, ZipWriter};
 
-fn generate_random_archive(size: usize) -> Vec<u8> {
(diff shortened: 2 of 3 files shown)


zamazan4ik:
@Plecra maybe will be better to avoid using `strum` here and construct an array of supported algorithms manually? Maybe smth like this https://github.com/kauhat/zip-rs/pull/1/files#diff-72de9ea1c97e4bb5074ece721c16c233aa29d6068cae340f32fb3801ba18c4aeR125 or anything else

Plecra:
Nice catch with the old dependencies. I think we might as well remove the benchmark code - the results aren't useful because of how non-deterministic they are, and we haven't given much of any focus to performance.

zamazan4ik:
@kauhat maybe you have some comments regarding this PR? Since this work is mostly based on your prev PR.

zamazan4ik:
I cannot say that the current benchmark code is useless. I think it could be used for future developments. E.g. for some changes, we really can check that we don't affect the performance a lot (or maybe even improved).

kauhat:
Hi, thanks for looking at this! I think I had made some progress and was looking into seeding Lorem text into a memory buffer before getting stuck somewhere and distracted by other things.

I planned to write a handful of benchmarks covering simple use-cases (e.g. mirroring the /examples) and test against all compression methods with a few sets of test data, for example:

- Many small files
- A few large files
- Lorem Ipsumm/text data
- Binary data

The goal was to have a place to compare the performance of the compressor implementations. Although results will be different per machine, the relative speeds between the compression methods should be somewhat comparable and we have some expectations of how fast they should be. So if new compression methods are added maybe it would be visible if one is performing uncharacteristically. I don't have much experience writing benchmarks though. 😄 

Criterion was nice to use and it looks like there are some [hooks we could plug into for profiling](https://www.jibbow.com/posts/criterion-flamegraphs/) if anyone needs to do that. Perhaps this would make issues like #243 easier to diagnose.

Potentially we could set up an action to run benchmarks on `master` commits so that we have some historical record of performance and can be notified of regressions. 

I'm aware that zip-rs has many users - so I don't want to add any extra dependencies to the main code path. Could Criterion and any others be made optional under a feature?

zamazan4ik:
Thanks for the intention to write some benchmarks!

> I'm aware that zip-rs has many users - so I don't want to add any extra dependencies to the main code path. Could Criterion and any others be made optional under a feature?

I don't think that it's problem - Criterion is `dev-dependency` so it shouldn't affect our users

Plecra:
I am not planning to work on performance of this version of the crate, as I am not pushing quickly towards the 0.10 release. The benchmarking strategies are sure to be very useful though, so we'll be reusing the changes made in this PR in the next PR on this subject

PR #271, opened 2022-01-24
Add SUPPORTED_COMPRESSION_METHODS constant

kauhat:
This PR adds a `zip::SUPPORTED_METHODS` const to get the implemented compression methods at runtime.

I couldn't figure out how to use the existing `src/compression.rs test::methods()` function from tests/, so exposing this could be useful. I'm not certain that it should be publically exposed though, so perhaps this should be `pub(crate)` instead?

I am not a pro Rust dev so I'm not sure if this PR is following best practices, however, Clippy doesn't see any problems.

Added for #272


kauhat:
Perhaps `CompressionMethod::iter()` would work better instead?

Plecra:
I suppose it's reasonable to make this a public API. It could be useful for users. I'd like to have `zip::{read, write}::SUPPORTED_COMPRESSION_METHODS`

zamazan4ik:
@kauhat can you update the PR regarding the @Plecra comment? Thanks!

kauhat:
I have updated the name to `SUPPORTED_COMPRESSION_METHODS`. It's passing all tests with recent changes from the master branch.

zamazan4ik (approved):

Issue #273, opened 2022-01-30
Break in AES PR


Plecra:
This is a repeat of #266, and the same solution is needed

zamazan4ik:
@Plecra can we close the issue since the related PR is already merged?

PR #274, opened 2022-01-30
remove `CompressionMethod::Aes` enum variant

Lireer:
If I understand the problem correctly #273 should be fixed by this.


zamazan4ik:
@Plecra is this solution fine for you?

Plecra:
Hm. I thought there were matches that needed removing but CI seems happy 🤔

Plecra (approved):
🤷 LGTM. Thanks :)

Issue #147, opened 2020-06-16
Laziness of ZipArchive and ZiipFile make for an awkward API


rylev:
Both `ZipArchive` and `ZipFile` are lazy, and only read from the underlying Reader when they absolutely have to. This is great for performance as zip archives can be very large, and we don't want to eagerly uncompress and read everything into memory. However, there are a few ways this makes the API a bit awkward:

### `ZipArchive::{by_name,by_index}` take `&mut self`
This is because these methods mutate the underlying reader. Ultimately this is not an ease fix as the laziness of both `ZipArchive` and `ZipFile` do not really allow us to change this to `&self` by using interior mutability (through use of something like `RefCell`). Most of the time having exclusive access of the ZipArchive is not an issue, but it can be confusing for newcomers who may not understand why the ZipArchive needs mutable/exclusive access for reading files.

### Iterator APIs are not supported
Currently the ZipArchive does not support iterators, requiring the user to get files by name or by index. Supporting iteration is not that easy as `iter()` normally takes `&self` and yielding `ZipFile`s from the iterator causes lifetime issues since `ZipFile`s are attached to the lifetime of the underlying reader. 

Question: what should we do about this? Can we keep the performance of the current API without some of these awkward parts?

Plecra:
I think we should be able to solve the first issue with better naming: building a `zip::Reader` from your `io::Read` would make it fairly intuitive that it needs similar access to `io::Read::read`.

There are a few options for a more intuitive API. I've been mocking up some alternatives, but they're basically all internal iterators. We'll probably end up with something like this:

```rust
for line in archive.flat_map(|file| Ok(if file.name.extension() == "log" {
    Some(io::BufReader::new(file).lines())
} else {
    None
})) {
  println!("{}", line);
}
```

When it's *eventually* stabilized, we should definitely take the opportunity to implement `StreamingIterator` too.

MaulingMonkey:
[...]

PR #276, opened 2022-02-05
Make `ZipArchive` cheap to clone

a1phyr:
This PR makes `ZipArchive` cheap to clone.

This enables the following pattern to read it from a shared reference: clone it and read from the clone. This works well if the reader is cheap to clone such as a `Cursor<Arc<[u8]>>` or a [`sync_file::SyncFile`](https://docs.rs/sync_file/latest/sync_file/struct.SyncFile.html).

Provides a good solution to #147

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -45,7 +45,9 @@ struct Shared {
 
 /// ZIP archive reader
 ///
-/// This type is cheap to clone if this is the case for the reader it uses.
+/// At the moment, this type is cheap to clone if this is the case for the
+/// reader it uses. However, this is not guaranteed by this crate and it may
+/// change in the future.
 ///
 /// ```no_run
 /// use std::io::prelude::*;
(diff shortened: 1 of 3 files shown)


zamazan4ik:
@Plecra can you review it please?

Plecra:
This is a reasonable addition, but `Clone`'s performance isn't going to be a hard guarantee. Let's not dig ourselves into a deeper compatibility corner :D

src/read.rs
```diff
@@ -33,8 +34,19 @@ mod ffi {
...
+struct Shared {
+    files: Vec<ZipFileData>,
+    names_map: HashMap<String, usize>,
+    offset: u64,
+    comment: Vec<u8>,
+}
+
 /// ZIP archive reader
 ///
+/// This type is cheap to clone if this is the case for the reader it uses.
```
Plecra, line 48: Please add that the performance of this may regress if you can ^^ We can merge either way

Plecra (approved):

Plecra (approved):

PR #278, opened 2022-02-28
Fix minimal versions

fujiapple852:
As of latest master (a9e143665526a40cd8e1c194085644f7ec58f953) the code does not compile with `minimal-versions` and the default features:

```shell
cargo +nightly update -Z minimal-versions && cargo +nightly check
```

This PR contains 3 commits which build on each other such that:
- The 1st makes the minimal update to the versions in `Cargo.toml` such that all tests pass (with `--all-features`)
- The 2nd specifies precise versions for all dependancies in Cargo.toml
- The 3rd makes the maximal (breaking change) update to the versions in `Cargo.toml` such that all tests pass (with `--all-features`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2018"
 
 [dependencies]
 aes = { version = "0.7.5", optional = true }
-byteorder = "1.3"
+byteorder = "1.3.4"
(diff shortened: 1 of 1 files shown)


zamazan4ik:
Thanks for the changes!

PR #282, opened 2022-03-19
implement lzma decompression

icewind1991:
Adds support for reading lzma compressed zip files.

Compression support is missing because the used lzma library doesn't implement writing (and I couldn't get any of the other lzma libraries working with the zip lzma header)


Plecra:
Your PR looks great, thankyou for the work on it 😊 Unfortunately, it looks like the LZMA library has gone unmaintained for 6 years now, and I can't take on maintenance for any bugs found in its implementation. As and when Rust has a maintained LZMA library, I'll be happy to come back to this

PR #285, opened 2022-03-21
Fix targets without 64-bit atomics.

### LINKED ISSUE #284: Restore MIPS support?
#276 removed MIPS (mips-unknown-linux-gnu) support. Is it possible to use smaller atomics to restore support for that platform? Or to disable this optimization for this target.

n3vu0r:
Fixes #284. As soon as `MSRV >= 1.60.0` we might use [`target_has_atomic = "64"`](https://github.com/rust-lang/rust/pull/94817/files) instead of `target_arch`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,9 @@ sha1 = {version = "0.10.1", optional = true }
 time = { version = "0.3.7", features = ["formatting", "macros" ], optional = true }
 zstd = { version = "0.10.0", optional = true }
 
+[target.'cfg(any(target_arch = "mips", target_arch = "powerpc"))'.dependencies]
+crossbeam-utils = "0.8.8"
+
 [dev-dependencies]
 bencher = "0.1.5"
 getrandom = "0.2.5"
--- a/src/read.rs
+++ b/src/read.rs
@@ -500,9 +500,9 @@ impl<R: Read + io::Seek> ZipArchive<R> {
     }
 
     /// Search for a file entry by name, decrypt with given password
-    /// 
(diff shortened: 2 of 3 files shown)


zamazan4ik:
Thanks for the PR! However, I think we cannot merge it yet since bumping MSRV to 1.60 violates our MSRV policy. I propose just wait for some time and then merge it.

If you need this fix urgently, I suggest just use your own fork during some time and then switch to the mainstream when it will be merged.

zamazan4ik:
Another option of course will be implementing missed functionality with some 3rd party creates which allows to enbale MIPS support without newer compiler.

paolobarbolini:
> Thanks for the PR! However, I think we cannot merge it yet since bumping MSRV to 1.60 violates our MSRV policy. I propose just wait for some time and then merge it.

This PR by itself doesn't bump MSRV. What we meant is that once the MSRV will be high enough we'll be able to replace `any(target_arch = "mips", target_arch = "powerpc")` with `target_has_atomic = "64"`

zamazan4ik:
Ah, I see - my bad. Just did not understand you correctly. LGTM. 

@Plecra could you check it too?

Plecra (approved):
Hm, I'm not such a fan of adding a dependency only for a platform we can't test, but this makes for a very tidy solution. Since it isn't public API, this doesn't have to be permanent.

Thanks for the fix!

messense:
Could you please release a new version?

zamazan4ik:
@Plecra that is a question for you )

Issue #212, opened 2020-11-19
Allow custom DEFLATE compression level


MCOfficer:
I'm compressing a massive text file using this crate, which takes a lot of time using `FileOptions::default()`. I experimented with 7z, and setting the compression level to 1 produces almost the same filesize, but is significantly faster.

However, It seems like the compression level can't be set using this crate. Is this correct?

Plecra:
That's right, we don't support compression configuration just yet. Our backwards compatibility guarantees prevent us from adding it to the current API, so it'll need to be properly added in the next release.

That said, this would at least be *possible* with [the raw reading API](https://github.com/zip-rs/zip/issues/205#issuecomment-727584125) we're adding. Would that be any help for you?

MCOfficer:
> 
> 
> That's right, we don't support compression configuration just yet. Our backwards compatibility guarantees prevent us from adding it to the current API, so it'll need to be properly added in the next release.
> 
> That said, this would at least be _possible_ with [the raw reading API](https://github.com/zip-rs/zip/issues/205#issuecomment-727584125) we're adding. Would that be any help for you?


It may be, but i feel like i'm not getting it. How would this help me with creating a zip file?

Plecra:
It doesn't 😅 I misunderstood what you were doing. Annoyingly, there's quite a lot of bookkeeping involved that I can't simply factor out. I don't expect this to be solved before the new year.

0ax1:
+1 on that this would be a great addition. Are there any updates on the status?

marcospb19:
+1, we would need this to solve https://github.com/ouch-org/ouch/issues/52.

@Plecra could you give a status update on this? How hard is it to implement? I'd probably be able to help by the end of this year depending on the difficulties involved.

marcospb19:
Does #286 closes this? @Plecra?

Plecra:
It sure does! These days, `writer.start_file("oopsies", FileOptions::default().compression_level(n)))` will let you write a file with the compression level you like. Thanks for the bump ^^

PR #286, opened 2022-03-22
Add support for specifying compression level

dmitryvk:
Hi,

I've been using zip-rs to compress somewhat large files on the fly.
Setting compression level from 6 (the default) to 1 makes a big difference in time to compress (about ~3-fold reduction).

In #212 some concerns regarding API compatibility were raised, but I don't see any concerns: field is added to ZipFileData which is not exported from the crate.

I also don't like using u32 to specify compression level (without range checking) - but it seems that flate2, bzip and zstd also use u32.

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -11,6 +11,7 @@ use std::default::Default;
 use std::io;
 use std::io::prelude::*;
 use std::mem;
+use std::ops::RangeInclusive;
 
 #[cfg(any(
     feature = "deflate",
@@ -103,6 +104,7 @@ struct ZipRawValues {
 #[derive(Copy, Clone)]
 pub struct FileOptions {
     compression_method: CompressionMethod,
+    compression_level: Option<i32>,
     last_modified_time: DateTime,
     permissions: Option<u32>,
     large_file: bool,
@@ -124,6 +126,7 @@ impl FileOptions {
                 feature = "deflate-zlib"
             )))]
             compression_method: CompressionMethod::Stored,
+            compression_level: None,
             #[cfg(feature = "time")]
             last_modified_time: DateTime::from_time(OffsetDateTime::now_utc()).unwrap_or_default(),
             #[cfg(not(feature = "time"))]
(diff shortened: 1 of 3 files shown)


src/write.rs
```diff
@@ -143,6 +151,20 @@ impl FileOptions {
...
     }
 
+    /// Set the compression level for the new file
+    ///
+    /// Range of values depends on compression method:
+    /// * `Deflated`: 0-9. `None` corresponds to 6
+    /// * `Bzip2`: 0-9. `None` corresponds to 6
+    /// * `Zstd`: 1-21. `None` corresponds to 3
+    /// * others: only `None` is allowed
+    /// `None` value specifies default compression level.
```
zamazan4ik, line 161: I suggest to leave the information about compression level only in this place and remove duplication in other places. It will reduce complication in future maintenance.

zamazan4ik (changes requested):

zamazan4ik:
Thanks for the PR! I am really glad to see this kind of functionality in the library.

Regarding range check question. Honestly, I also would like to see some kind of range checking. Maybe we can at least provide to the user some kind of helper. But in this case, we will introduce some additional kind of coupling between our library and dependent crates (we will need our range checkers with their supported ranges for each algorithm). I cannot say right now, which way is better.

src/write.rs
dmitryvk: Agreed; removed other doc comments

dmitryvk:
I've updated the PR.
Extra doc comments are removed.
I've added range checking for compression level. Values for deflate and bzip2 seem to be standardized, but zstd is somewhat complicated.
It seems that over time compression level range of zstd was increased: zstd crate mentions levels 1-21 and zstd docs state that maximum is 22 and negative values are allowed. Wikipedia states that -7 is the lowest value. Looking at zstd source code it seems that the actual limit is much lower (seems to be -128K though I didn't 100% verify it). Relaxing the constraints is easier so I went with more conservative range of -7 to 22.

dmitryvk:
> But in this case, we will introduce some additional kind of coupling between our library and dependent crates

Yes, this is a coupling, but it's already present. Also, all of the crates provide means to query for allowed compression level values (`flate2::Compression::none/best`, `bzip2::Compression::none/best`, `zstd::compression_level_range`)

zamazan4ik (approved):

zamazan4ik:
> Yes, this is a coupling, but it's already present. Also, all of the crates provide means to query for allowed compression level values (flate2::Compression::none/best, bzip2::Compression::none/best, zstd::compression_level_range)

Yeah, that is what I wanted to see - using for the range checks ranges directly from our dependencies and does not introduce such ranges in our codebase. 

LGTM.

Since this contribution is quite important from the feature perspective, I suggest wait for some time for the review from @Plecra

Plecra (approved):
I like it :) Thanks for the changes

dmitryvk:
Thanks, I didn't expect such quick response:)

PR #288, opened 2022-03-25
Fix ZIP64 write support.

### LINKED ISSUE #287: ZIP64 spec violation
>4.5.3 -Zip64 Extended Information Extra Field (0x0001):  
>  
>The following is the layout of the zip64 extended   
>information "extra" block. If one of the size or  
>offset fields in the Local or Central directory  
>record is too small to hold the required data,  
>a Zip64 extended information record is created.  
>The order of the fields in the zip64 extended   
>information record is fixed, **but the fields MUST  
>only appear if the corresponding Local or Central  
>directory record field is set to 0xFFFF or 0xFFFFFFFF.**  

Code:

```rust

use std::fs::OpenOptions;
use std::io::{BufWriter, Write};

use anyhow::Result;
use zip::write::FileOptions;
use zip::ZipWriter;

fn main() -> Result<()> {
    let outfile = OpenOptions::new().create(true).truncate(true).write(true).open("test.zip")?;
    let mut bw = BufWriter::new(ZipWriter::new(outfile));
    const SIZE: u64 = 4294967297;
    bw.get_mut().start_file("testfile", FileOptions::default().large_file(true))?;
    for _ in 0..SIZE {
        bw.write_all(&[0])?;
    }
    bw.flush()?;
    bw.get_mut().finish()?;
    Ok(())
}

```

The local record of `test.zip` lists `testfile`'s compressed size as `6A A7 3F 00` (4171626) in **both** the standard local record and ZIP64 extended information record, which is incorrect. (If you trust 7-Zip, it'll show "headers error" warning for this, and no warnings if you overwrite the local csize field with `FF`s)

Options I know of to fix this:

1. Always set the local record compressed and uncompressed sizes to `0xFFFFFFFF` when `large_file` is `true`, and only specify their real values in the ZIP64 record.
2. Allow users to give the size of the file in advance, and if it's below a certain size don't write the ZIP64 csize field and rely on the local record's csize field. Maybe <3GB? I don't think any compression method will manage to add 1GB to even a fully random 3GB file. (DEFLATE seems to add only a few MB). This is basically an optimization for option 1.
3. Re-write the entire ZIP file after finishing in order to minimize its size (comedy option)

n3vu0r:
Fixes #287 and raw copying of large files.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -727,17 +727,17 @@ fn parse_extra_field(file: &mut ZipFileData) -> ZipResult<()> {
         match kind {
             // Zip64 extended information extra field
             0x0001 => {
-                if file.uncompressed_size == 0xFFFFFFFF {
+                if file.uncompressed_size == spec::ZIP64_BYTES_THR {
                     file.large_file = true;
                     file.uncompressed_size = reader.read_u64::<LittleEndian>()?;
                     len_left -= 8;
                 }
-                if file.compressed_size == 0xFFFFFFFF {
+                if file.compressed_size == spec::ZIP64_BYTES_THR {
                     file.large_file = true;
                     file.compressed_size = reader.read_u64::<LittleEndian>()?;
                     len_left -= 8;
                 }
-                if file.header_start == 0xFFFFFFFF {
+                if file.header_start == spec::ZIP64_BYTES_THR {
                     file.header_start = reader.read_u64::<LittleEndian>()?;
                     len_left -= 8;
                 }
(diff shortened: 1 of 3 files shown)


zamazan4ik:
Thank you for the PR. I have some suggestions how it can be improved.

1. I suggest introducing some dedicated constant name to `0xFFFFFFFF` variable and reuse it.
2. Could you add some tests (maybe unit tests?) which covers this change? It will help us with the future maintenance

Everything else looks fine!

n3vu0r:
Yes, I should add some tests as part of this PR. Also `u32::MAX` might do the trick.

zamazan4ik:
> Also u32::MAX might do the trick.

Even with this predefined constant you can name somehow this constant in our code which will have some meaningful name according to ZIP spec (or just reference the ZIP spec as a comment to this constant)

zamazan4ik:
@n3vu0r Just ping me, then the PR will be ready for the review :)

n3vu0r:
@zamazan4ik Adding tests for writing large files is a bit more involved. Feel free to review/merge this PR.

In a separate PR we can reuse `Zip64File` in `tests/zip64_large.rs` by also implementing `Write` for it. Then call `raw_copy_file` and compare subslices of the different byte blocks. The blocks will not completely equal as only the ZIP64 extra field is preserved and there are other extra fields present. Also large numbers of files in the central directory should be covered by a test.

zamazan4ik:
Thanks!

Issue #248, opened 2021-08-06
Can we expose `CentralDirectoryEnd::find_and_parse`?


jooyunghan:
I found that inspecting a zip file can benefit from `CentralDirectoryEnd::find_and_parse` since everything starts from there. Can we expose the API as public?

Plecra:
Sure can! I want to put this in a subcrate which will expose the zip algorithms with a less stable API. Refactoring the current implementation into an in-tree dependency would be perfect

jooyunghan:
Sounds great! Thank you.

zamazan4ik:
@Plecra do you have any updates on this?

Plecra:
Moving the internal parsing apis into a subcrate (`zip-format`?) would be a perfectly solid solution. It should be unstable, as in pre-1.0, and should make it cleaar that version bumps are common because `zip` is the stable API. Maybe at some point it could be stabilized once the core crate itself is more secure.

Plecra:
This is WIP in #292 and kinda subsumed by #58. Lets call this "closing as duplicate" :D

PR #290, opened 2022-03-25
Add `zip::spec::{CentralDirectoryHeader, DataDescriptor, LocalFileHea…

aweinstock314:
…der}`, and partially integrate them into `zip::{read, write}`.

These are nice to have for manipulating zip files at a low level (e.g. compacting data descriptors into local file headers to save space, creating orphaned local file headers not present in the central directory containing custom metadata).

This PR doesn't make the `spec` mod public yet, but the hope is that these can be exposed by a subcrate similar to #248.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -640,70 +640,40 @@ pub(crate) fn central_header_to_zip_file<R: Read + io::Seek>(
     archive_offset: u64,
 ) -> ZipResult<ZipFileData> {
     let central_header_start = reader.seek(io::SeekFrom::Current(0))?;
-    // Parse central header
-    let signature = reader.read_u32::<LittleEndian>()?;
-    if signature != spec::CENTRAL_DIRECTORY_HEADER_SIGNATURE {
-        return Err(ZipError::InvalidArchive("Invalid Central Directory header"));
-    }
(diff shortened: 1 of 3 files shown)


aweinstock314:
> As far as I remember, @Plecra had some concerns about the similar designs - he wanted to have such functionality in some external crate. However, maybe as a start point, it could be acceptable from my point of view to have such functionality in this crate.

If I understand correctly, the reason to expose low-level details as a separate crate is to avoid committing to non-final designs as part of zip's public API. I agree with this reasoning, which is why this PR does not extend zip's public API (i.e. it does not change `mod spec;` to `pub mod spec;` in `src/lib.rs`).

zamazan4ik (approved):

zamazan4ik:
@aweinstock314 could you please rebase your changes on the latest master?

Also, would be awesome, if you add some unit-tests for the newly introduced classes :) Anyway, I like the current code much more since it is more readable and maintainable :)

aweinstock314:
I've rebased. I'll add some tests that assert that the serialization round-trips properly.

aweinstock314:
I've added tests.

zamazan4ik (approved):

zamazan4ik:
Thanks!

Issue #291, opened 2022-03-25
Broken ZIP files created by commit a1da686


AlyoshaVasilieva:
PR #290 (rev a1da6860ea423e999b731fca866ed54505e0096e) appears to have broken writing, at least header writing. Creating test.zip by my example code in #287, tools report:

7-Zip: 

```
Error: Headers Error
Unconfirmed start of archive

Warning: Headers Error
There are some data after the end of the payload data
```

`zip -T test.zip` (zip 3.0-11build1 on Ubuntu 20.04):

```
  error:  invalid compressed data to inflate testfile
test of test.zip FAILED

zip error: Zip file invalid, could not spawn unzip, or wrong unzip (original files unmodified)
```

Neither 7-Zip nor `unzip` can extract the archive created by this library.

The same code, when using rev 574bb3df1748a89c8cb1dbed65e625526e4496ee, the immediately preceding commit, creates a working ZIP file.

zamazan4ik:
@aweinstock314 check the issue please

zamazan4ik:
@AlyoshaVasilieva I am sorry for the broken master branch. At least it shows that now we have lack of tests

zamazan4ik:
If we do not fix the issue quickly - I will revert the corresponding PR.

AlyoshaVasilieva:
No worries about broken master, better now than released.

I'm not good at writing CI but I'd recommend maybe creating test files and using `zip -T` command on a Linux runner to test them as a possible improvement to testing. It won't find everything (e.g. I think it doesn't care about wrong local headers if central record is fine) but it would detect broken files like here.

n3vu0r:
I'm on it.

zamazan4ik:
I have reverted the changes in master branch. So if you want to continue the work - please rebase :)

n3vu0r:
Does the fix as part of #292 pass your tests?

zamazan4ik:
@AlyoshaVasilieva could you confirm please that with the latest release the issue is resolved? Thanks!

AlyoshaVasilieva:
I believe it was fixed, will test again once I wake up

AlyoshaVasilieva:
The test I've been using works fine with 0.6.2. Also works fine with the rejected PR #292.

Issue #228, opened 2021-03-14
Extraction logic is vulnerable to Zip Bomb attacks


zamazan4ik:
Hi!

Thank you for the library!

I have a question: how this library handles zip bombs. Is there any built-in protection? If no, can you suggest an option, how can I handle this?

My use case is: the user can upload a zip file into the system, then the system shall extract some information from the archive and store it somewhere. As you see, the user can try to send some malformed archives and crash my system.

Thanks!

Plecra:
My understanding of zip bombs is that they specifically target antivirus software by appearing to contain zip files which contain zip files, and so on. This can crash an antivirus that attempts to recursively extract them, but shouldn't have an affect on any of `zip`s APIs as they are non recursive.

Are you aware of a method to attack the current implementation of `extract`?

ZaMaZaN4iK:
No, I'm not aware of any kind of already existing method to attack the current implementation (actually I didn't try any).

About zip-bombs. Not only antivirus. E.g. recursive zip bombs can just cause DoS during the extraction process. Why I am explicitly asking about such things - because I remember already existing similar CVE in other ZIP libraries. E.g this one: https://bugs.python.org/issue36462 (yeah, that's not Rust and possibly that's attack is not applicable to zip-rs - just an example).

Plecra:
Thanks for the link :) My first impression is a bit tangential to the `zip` library's threat model, and we provide the apis for a user to inspect the size of a zip file. Documenting the potential risk of writing untrusted data to the filesystem is a good idea.

For that, I'd like to have a clearer picture of the potential of this kind of attack...

- `compressed_size:uncompressed_size` ratio
- Duplicate file entries
- Anything else?

If we can reliably identify 'unreasonable' zip files, then I'd also be happy to include those checks in archive construction.

Plecra:
To answer your initial question, there is no protection in place at the moment and you can defend against this kind of attack by setting reasonable upper limits on the sizes of the files in the archives your users upload. 

The only reason we don't do this by default is that it's trickier to determine what is 'reasonable' for all applications.

ZaMaZaN4iK:
Thanks for the answer. Let's left the issue open. Maybe we will do some zip attacks analysis later (I'm not a professional in this area) and at least document some possible ways to prevent the most dangerous attack vectors.

zamazan4ik:
Just found a PR with a bunch of Zip related security fixes: https://github.com/rust-compress/rc-zip/pull/18

[...]

PR #292, opened 2022-03-26
Add spec structs

n3vu0r:
Second try for #290 hopefully fixing #291.

I haven't reviewed all of the original changes yet but also haven't seen any problems so far.

(I've manually set the author of the original changes to @aweinstock314 with a separate commit as I wasn't able to rebase.)

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -640,70 +640,43 @@ pub(crate) fn central_header_to_zip_file<R: Read + io::Seek>(
     archive_offset: u64,
 ) -> ZipResult<ZipFileData> {
     let central_header_start = reader.seek(io::SeekFrom::Current(0))?;
-    // Parse central header
-    let signature = reader.read_u32::<LittleEndian>()?;
-    if signature != spec::CENTRAL_DIRECTORY_HEADER_SIGNATURE {
-        return Err(ZipError::InvalidArchive("Invalid Central Directory header"));
-    }
+    let central_header = spec::CentralDirectoryHeader::parse(reader)?;
 
-    let version_made_by = reader.read_u16::<LittleEndian>()?;
-    let _version_to_extract = reader.read_u16::<LittleEndian>()?;
-    let flags = reader.read_u16::<LittleEndian>()?;
-    let encrypted = flags & 1 == 1;
-    let is_utf8 = flags & (1 << 11) != 0;
-    let using_data_descriptor = flags & (1 << 3) != 0;
-    let compression_method = reader.read_u16::<LittleEndian>()?;
-    let last_mod_time = reader.read_u16::<LittleEndian>()?;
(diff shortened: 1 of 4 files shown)


aweinstock314:
Looks good to me (assuming 614f55490e057ad67ae0cc241b3df604e7b633d6 is a revert of 77294fd3ce4d9657a31ea07894ca6fb3e83c2174). I probably mishandled the rebase of #288's commits for #290.

Would you be willing to add unit tests for the zip64 support (not necessarily as part of this PR)?

n3vu0r:
Yes, it's supposed to revert the revert. I'm planning to add an exhaustive ZIP64 test and some comments to the code (not necessarily as part of this PR). Not sure when, maybe next weekend.

---

Btw, I ran into exactly this as well, so I mention it. This screwed up the header and resulted in CRC-32 mismatches:

```rust
let mut zip64_extra_field = vec![0; 20];
write_local_zip64_extra_field(&mut zip64_extra_field, file)?;
```

It doesn't overwrite the 20 bytes of zeros as intended but instead extends them by the extra field. This would do the trick:

```rust
let mut zip64_extra_field = vec![0; 20];
write_local_zip64_extra_field(&mut &mut zip64_extra_field[..], file)?;
```

It's `impl Write for Vec<u8>` vs `impl Write for &mut [u8]`. Former extends, latter overwrites.

Plecra:
These structures are not going to be exposed publicly as-is, and I see no need for the changes in the current version. Thanks for the work you've done on this, but we will make these changes as and when we make the API public, so that we can properly motivate the design of them.

(If you think I'm missing something feel free to say ^^)

aweinstock314:
Even if they're not exposed yet, having the specification structures separated out makes the code clearer (currently, the code mixes concerns between reading/writing the structures as they appear in the file and casting them into an intermediate representation that supports deferred lookups and writes in a way that's agnostic between a few different equivalent ways to represent the same metadata (e.g. data descriptors)), and adds unit tests for the specification structures themselves.

These structures suffice for adding custom metadata via an orphaned LFH (e.g. https://github.com/aweinstock314/zip/blob/8acc05175b11aa372bc5f11d5351e3ce0885c50b/examples/reorder_cd.rs, which makes retrieving the central directory via HTTP range requests more efficient, since some servers give worse response times for range requests with offsets from the end of a file), and I'm confident that they'll generalize to crafting other sorts of files (e.g. overlapping LFHs with file contents to write unit tests for #228).

I'd like these to be exposed as part of the proposed `zip-format` crate, and would like them to be merged as private types prior to that to avoid needing to rebase around other conflicting changes to reading/writing these structures.

Plecra:
This was and still is a good idea, whoops! The definitions look great, and I'll bring them in as `unstable` API ASAP

Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/71.

PR #302, opened 2022-04-11
Use Seek.stream_position instead of seek(0).

PJB3005:
BufReader can cache the result of stream_position() so it's fast, whereas seek(0) discards the read buffer and passes through straight to the OS. This drastically worsens the efficiency of loading performance when using BufReader (or anything else to avoid all these tiny reads going straight to the kernel).

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -639,7 +639,7 @@ pub(crate) fn central_header_to_zip_file<R: Read + io::Seek>(
     reader: &mut R,
     archive_offset: u64,
 ) -> ZipResult<ZipFileData> {
-    let central_header_start = reader.seek(io::SeekFrom::Current(0))?;
+    let central_header_start = reader.stream_position()?;
     // Parse central header
     let signature = reader.read_u32::<LittleEndian>()?;
     if signature != spec::CENTRAL_DIRECTORY_HEADER_SIGNATURE {
--- a/src/write.rs
+++ b/src/write.rs
@@ -348,7 +348,7 @@ impl<W: Write + io::Seek> ZipWriter<W> {
 
         {
             let writer = self.inner.get_plain();
-            let header_start = writer.seek(io::SeekFrom::Current(0))?;
+            let header_start = writer.stream_position()?;
 
             let permissions = options.permissions.unwrap_or(0o100644);
             let mut file = ZipFileData {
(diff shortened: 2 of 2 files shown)


zamazan4ik:
I am a little wondered why in previous versions, `seek(0)` was used. Seems really strange to me

PJB3005:
`stream_position()` wasn't added until Rust 1.51, to my massive surprise. So that's probably why.

zamazan4ik:
Thanks for the PR!

PJB3005:
I don't have the most scientific benchmark setup here because I just ran some ETW profiles when I noticed my zip files (like ~50 MB, ~13k mostly-tiny files) took way longer to open than they should on debug mode, then cross-referenced the code with the amount of seek calls and yep.

I just made a quick and dirty Read + Seek impl that doesn't pass through `io::SeekFrom::Current(0)` to the base stream:
```rust
struct FTellStream<R> {
    inner: R,
}

impl<R: Read + Seek> FTellStream<R> {
    pub fn new(inner: R) -> Self {
        FTellStream { inner }
    }
}

impl<R: Read + Seek> Read for FTellStream<R> {
    fn read(&mut self, buf: &mut [u8]) -> Result<usize, io::Error> {
        self.inner.read(buf)
    }

    fn read_exact(&mut self, buf: &mut [u8]) -> io::Result<()> {
        self.inner.read_exact(buf)
    }
}

impl<R: Read + Seek> Seek for FTellStream<R> {
    fn seek(&mut self, from: io::SeekFrom) -> Result<u64, io::Error> {
        if from == io::SeekFrom::Current(0) {
            return self.inner.stream_position();
        }

        self.inner.seek(from)
    }
}
```

Without, doing this on these zip files:
```rust
        let a = Instant::now();
        let file = fs::File::open(zip_path).unwrap();
        let mut zip = ZipArchive::new( BufReader::new(file)).unwrap();
        let b = Instant::now();

        println!("{}, {:?}", zip.len(), b.duration_since(a));
```

I get like:

```
17092, 85.636ms
17092, 86.8534ms
17092, 84.1522ms
17076, 82.9949ms
17092, 84.8181ms
17092, 84.4906ms
```

With:

```rust
        let a = Instant::now();
        let file = fs::File::open(zip_path).unwrap();
        let mut zip = ZipArchive::new(FTellStream::new(BufReader::new(file))).unwrap();
        let b = Instant::now();

        println!("{}, {:?}", zip.len(), b.duration_since(a));
```

```
17092, 38.4624ms
17092, 38.2349ms
17092, 40.5351ms
17076, 38.3368ms
17092, 37.9675ms
17092, 37.6783ms
```

Should be noted that this is on Windows, for however much that matters.

PR #303, opened 2022-04-11
Use some ::with_capacity when reading zip file.

PJB3005:
Now with a proper benchmark:

```
before:
test read_metadata ... bench:  11,988,380 ns/iter (+/- 1,411,339)
after:
test read_metadata ... bench:   9,600,650 ns/iter (+/- 804,604)
```

At this point most of the cost is memory allocations and hashmap inserts. You'd need a two-pass parser (one to get all file sizes to pre-allocate large buffers for the variable-length data, another to fill them) and that's far more complexity than I'd want to put effort in.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,3 +42,7 @@ default = ["aes-crypto", "bzip2", "deflate", "time", "zstd"]
 [[bench]]
 name = "read_entry"
 harness = false
+
+[[bench]]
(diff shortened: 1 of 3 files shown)


zamazan4ik:
Could you please reformat your code?

PJB3005:
Just did

zamazan4ik (approved):

PR #306, opened 2022-04-16
Annotate why some crypto dependencies can't be upgraded

indygreg:
I found myself in dependency hell related to a bunch of RustCrypto
crates and have been chasing down intermediate dependencies.

It looks like we can't upgrade these dependencies in zip-rs because
of the current 1.54.0 MSRV. I figured I'd leave comments in the
Cargo.toml so others are aware.

While I'm here, is there any appetite in bumping the MSRV to 1.56 so we
can upgrade these crates? It can be exceptionally difficult to adopt the latest
versions of crates under the RustCrypto umbrella because some of these crates
are very widely used. I could probably code up a PR. Let me know...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,16 @@ Library to support the reading and writing of zip files.
 edition = "2018"
 
 [dependencies]
+# 0.8 requires 2021 edition and MSRV >=1.56.
 aes = { version = "0.7.5", optional = true }
 byteorder = "1.4.3"
 bzip2 = { version = "0.4.3", optional = true }
+# 0.2 requires 2021 edition and MSRV >=1.56.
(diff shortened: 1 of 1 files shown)


zamazan4ik:
@indygreg since our MSRV prevents updating our dependencies AND our MSRV policy says about support only about 4 latest minor versions, I suggest you just create a PR with bumping a MSRV version to 1.57 (1.57, 1.58, 1.59, 1.60 - is fine).

For this PR you should:
* Update the README about the cuurent MSRV
* Update CI scripts
* Update dependencies as you want

indygreg:
Sorry about that - I got distracted with... life. Feel free to close this PR if you agree that it is superseded by #310.

PR #307, opened 2022-04-25
Add read fuzzing module

nickbabcock:
As someone who has personal projects that take untrusted zips as input,
it is important to me to be able to fuzz the zip project to simulate
possible inputs and to ensure the projects are not vulnerable.

This commit adds a cargo fuzz module for reading and extracting input.

The `fuzz` directory was scaffolded with a `cargo fuzz init`

I added a CI step to guard against the fuzz module decaying over time.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -74,3 +74,19 @@ jobs:
 
     - name: Docs
       run: cargo doc
+
+  fuzz:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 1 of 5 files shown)


zamazan4ik:
I like the idea since I am also a "fuzz fanboy" :) By the way, we are already integrated into the Google Oss Fuzz project: https://github.com/google/oss-fuzz/tree/master/projects/zip-rs

And if you have any ideas, how it can be improved - you are welcome. It really will help us.

zamazan4ik (approved):

zamazan4ik:
Just an idea: we can put a note about fuzz testing into our README somehow. If you would like to do it - I suggest do it in this PR. Otherwise, I will merge it as is.

zamazan4ik (approved):

PR #308, opened 2022-04-25
Fix capacity overflow on invalid zips reads

nickbabcock:
Preemptively allocating structures with the number of reported files can
lead to trouble as an invalid zip can still have a valid central
directory end that is fed into a `with_capacity` causing it to overflow.

This commit introduces a heuristic that will use the reported number of
files as long as the number is less than the cde offset.

Benchmarks were unaffected by this change.

Found by #307


zamazan4ik:
Hi. Thanks for the PR! Could you please escribe the difference between two test archives? I am just interested,  why do we need two test cases here.

Other things look fine!

nickbabcock:
I originally wrote the following conditional

```rust
let file_capacity = if number_of_files > cde_start_pos as usize {
    0
} else {
    number_of_files
};
```

as: 

```rust
let file_capacity = if number_of_files > footer.central_directory_offset as usize {
    0
} else {
    number_of_files
};
```

Before realizing that `footer.central_directory_offset` is also user supplied. The second test case sneaks past the original conditional

Not sure if you see too much benefit there, but I can reduce the number of test cases to just the second one if you want.

zamazan4ik:
No-no, it is fine to have both cases in this situation since they are semantically different. In this case I just suggest you to rename test cases somehow for better explanation, why and where they are different. Maybe place some comments about it inside the tests cases too.

nickbabcock:
Updated 👍

zamazan4ik (approved):

PR #311, opened 2022-05-01
chore: update dependencies (pbkdf2 0.11.0, flate2 1.0.23)

striezel:
This is a follow up to #310 and the now closed #306.

It gets the dependencies closer to what #306 _presumably_ would have been, had the MSRV already been at 1.57 when that PR was opened.


joehoyle:
Did this make its way into a release?

zamazan4ik:
I do not think so. The recommended way to use the library - from the master branch. I have no permission to release the library.

PR #320, opened 2022-09-12
Depend Fix [zstd]

wyatt-herkamp:
when compiling with Actix web the exact version depend 0.11.0 creates a conflict with actix-web

By setting it to just 0.11 it can use any patch version of 0.11 so 0.11.0 or 0.11... 

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        rust: [stable, 1.57.0]
+        rust: [stable, 1.59.0]
 
     steps:
     - uses: actions/checkout@master
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ hmac = { version = "0.12.1", optional = true, features = ["reset"] }
 pbkdf2 = {version = "0.11.0", optional = true }
 sha1 = {version = "0.10.1", optional = true }
 time = { version = "0.3.7", features = ["formatting", "macros" ], optional = true }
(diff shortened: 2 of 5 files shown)


danylaporte:
I have run into the conflict version with actix-web. Is it possible to fix the clippy issues and merge the pull request?

wyatt-herkamp:
> I have run into the conflict version with actix-web. Is it possible to fix the clippy issues and merge the pull request?

So the clippy breaks are just related to new clippy rules nothing I changed. However they should be fixed now

Heliozoa:
` zstd = { version = "0.11", optional = true }` and ` zstd = { version = "0.11.0", optional = true }` are the exact same from Cargo's perspective because Cargo (mostly) follows [SemVer](https://semver.org/). `0.11.0` already allows use of any `0.11.z` patch release.

The problem is that the latest *release* of zip, 0.6.2, still uses zstd 0.10: https://crates.io/crates/zip/0.6.2/dependencies

wyatt-herkamp:
> ` zstd = { version = "0.11", optional = true }` and ` zstd = { version = "0.11.0", optional = true }` are the exact same from Cargo's perspective because Cargo (mostly) follows [SemVer](https://semver.org/). `0.11.0` already allows use of any `0.11.z` patch release.
> 
> The problem is that the latest _release_ of zip, 0.6.2, still uses zstd 0.10: [crates.io/crates/zip/0.6.2/dependencies](https://crates.io/crates/zip/0.6.2/dependencies)

That's what I had thought. I will revert that file change and keep the pr open as a clippy warning fix

PR #321, opened 2022-10-02
Add zstd-pkg-config feature

figsoda:
This is so consumers of `zip` can use the `pkg-config` feature of `zstd` without manually adding a `zstd` dependency


Plecra:
I believe the intended use of cargo is for dependent crates to explicitly opt into zstd themselves - this allows us to potentially update to newer versions of zstd internally even if they remove pkg-config support. If you manually add a zstd dependency, cargo will automatically find the suitable version of zip to continue supporting pkg-config. Do you think I'm getting something wrong here?

PR #330, opened 2022-11-18
Implement TryFrom<OffsetDateTime> for DateTime

aquacash5:
This allows users to use try_into when converting `OffsetDateTime` into `DateTime` and deprecates the previous specialized method.

### DIFF
--- a/src/result.rs
+++ b/src/result.rs
@@ -84,17 +84,15 @@ impl From<ZipError> for io::Error {
 
 /// Error type for time parsing
 #[derive(Debug)]
-pub enum ZipDateTimeError {
-    /// The date was on or before 1980, or on or after 2107
-    DateTimeOutOfBounds,
-}
+pub struct DateTimeRangeError;
 
-impl fmt::Display for ZipDateTimeError {
+impl fmt::Display for DateTimeRangeError {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            ZipDateTimeError::DateTimeOutOfBounds => write!(fmt, "datetime out of bounds"),
-        }
+        write!(
+            fmt,
+            "a date could not be represented within the bounds the MS-DOS date range (1980-2107)"
+        )
     }
 }
 
-impl Error for ZipDateTimeError {}
+impl Error for DateTimeRangeError {}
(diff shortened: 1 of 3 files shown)


src/result.rs
```diff
@@ -81,3 +81,20 @@ impl From<ZipError> for io::Error {
         io::Error::new(io::ErrorKind::Other, err)
     }
 }
+
+/// Error type for time parsing
+#[derive(Debug)]
+pub enum ZipDateTimeError {
```
Plecra, line 87: Please rename this to `DateTimeRangeError` and make it a unit struct

src/result.rs
```diff
@@ -81,3 +81,20 @@ impl From<ZipError> for io::Error {
...
+    /// The date was on or before 1980, or on or after 2107
+    DateTimeOutOfBounds,
+}
+
+impl fmt::Display for ZipDateTimeError {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            ZipDateTimeError::DateTimeOutOfBounds => write!(fmt, "datetime out of bounds"),
```
Plecra, line 95: "a date could not be represented within the bounds the MS-DOS date range (1980-2107)"

src/types.rs
```diff
@@ -167,6 +171,7 @@ impl DateTime {
     ///
     /// Returns `Err` when this object is out of bounds
     #[allow(clippy::result_unit_err)]
+    #[deprecated(note = "use `DateTime::try_from()`")]
     pub fn from_time(dt: OffsetDateTime) -> Result<DateTime, ()> {
         if dt.year() >= 1980 && dt.year() <= 2107 {
```
Plecra, line 176: Use `DateTime::try_from` within this method so we reuse the logic

Plecra (changes requested):
This is great, thanks! It only makes sense with `Try*` in the prelude these days.

Plecra:
@zamazan4ik would like to get this in and then publish 0.6.4. If you get the chance to prep the release before me, that'd be fantastic!

aquacash5:
@Plecra These should address the changes you requested.

Plecra (approved):
Thanks!

Plecra:
Need to run the tests on CI - could you sort out those warnings?

aquacash5:
That should fix it. I tried running the commands in the workflow on all the computers I have access to. I don't get any warnings or errors.

Plecra:
Alright, just rebased that on the clippy fixes :) I think we're left with a couple deprecated references which I can fix up just now...

PR #333, opened 2023-01-03
Disable `time` features

lnicola:
I left these enabled in `dev-dependencies`, so that the tests still work, but to be honest, I don't think those tests are really useful.

The reason to disable this is the proc macro stuff has a significant build time impact, and it's not really something `zip` users need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+
+## [0.6.4] - unreleased
+
+### Changed
+
+ - [#333](https://github.com/zip-rs/zip/pull/333): disabled the default features of the `time` dependency, and also `formatting` and `macros`, as they were enabled by mistake.
+ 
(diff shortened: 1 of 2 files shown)


zamazan4ik:
@Plecra we need your opinion here.

Plecra (changes requested):
Argh, more compatibility hazards. Fine, it's worth it. Hopefully we won't run into compatibility hazards, though I'd like to use a changelog to help users upgrading to the new version.

Plecra:
Hmm, I'm not sure about standard ways to to this, but I can pop a `CHANGELOG.md` in the root.

Plecra (approved):

Plecra:
Oops! We've got a bug in the fuzz CI. You'll need to add `"std"` to the featureset for `time`, otherwise we can't get the current time.

Plecra:
(That's https://docs.rs/time/latest/time/struct.OffsetDateTime.html#method.now_utc btw - it mentions the feature dep)

Plecra:
We need it for release `dependencies` too :) the `now_utc` API is used for writing the current time in file metadata.

lnicola:
Ah, okay. Fixed.

Plecra (approved):
LGTM! Thanks a bunch

lnicola:
Thanks! Do you think you'll cut out a new release soon?

Plecra:
Just after addressing the CVE issue, yep!

PR #334, opened 2023-01-06
Deletion of zip entries in-place

### LINKED ISSUE #283: Deletion of zip entries in-place
Rewriting will take way too long on mobile devices, so deleting in-place would be appreciated.

baloo:
fixes #283 

This brings the changes made by @rushiiMachine to support in-place deletion.

I tried my best  to keep original authorship, but I fixed a couple tests, documentation and formatting. I'm also adding a test for file deletion on top of @rushiiMachine 's work.

### DIFF
--- a/examples/write_dir.rs
+++ b/examples/write_dir.rs
@@ -1,7 +1,7 @@
 use std::io::prelude::*;
-use std::io::{Seek, Write};
 use std::iter::Iterator;
 use zip::result::ZipError;
+use zip::truncate::Truncate;
 use zip::write::FileOptions;
 
 use std::fs::File;
@@ -69,7 +69,7 @@ fn zip_dir<T>(
     method: zip::CompressionMethod,
 ) -> zip::result::ZipResult<()>
 where
-    T: Write + Seek,
+    T: Write + Seek + Truncate + Read,
 {
     let mut zip = zip::ZipWriter::new(writer);
     let options = FileOptions::default()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,3 @@
-#![feature(seek_stream_len)]
 //! A library for reading and writing ZIP archives.
 //! ZIP is a format designed for cross-platform file "archiving".
 //! That is, storing a collection of files in a single datastream
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![feature(seek_stream_len)]
 //! A library for reading and writing ZIP archives.
 //! ZIP is a format designed for cross-platform file "archiving".
 //! That is, storing a collection of files in a single datastream
@@ -39,6 +40,7 @@ mod crc32;
 pub mod read;
 pub mod result;
 mod spec;
+pub mod truncate;
 mod types;
 pub mod write;
 mod zipcrypto;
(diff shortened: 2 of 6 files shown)


rushiiMachine:
Probably should add a test for delete with fill_void
Also, I didn't bother adding but you should probably replace the file name with a placeholder (maintaining equal byte size) when deleting with fill_void

baloo:
If I read the code correctly, which I may not.
When we remove with fill void, we will remove the file from the list for files (the header). And on `finalize()` we will completely rewrite the central directory header. I'm not sure that is necessary to replace the filename. So I don't think it's necessary to overwrite this with a placeholder, or that we actually want that.

rushiiMachine:
> I'm not sure that is necessary to replace the filename

The point of this https://github.com/zip-rs/zip/pull/334/files#diff-8e31462ea4b11fdadc15fd98e895c1902595d9d34947c58eecca1125a5e6224dR902-R905 is to not retain potentially sensitive entry data.
The same logic could be applied to filename, which could be sensitive too. It should be erased so that it doesn't leak.

Plecra:
Heya! Thankyou for sending this in. As rushii mentioned before, this API isn't appropriate for the public API of the crate :) If the goal is to overwrite the space in a zip archive, recreating the archive from scratch is more suitable.

@rushiiMachine I see one of your commits is fixing a corruption of the archive. Can you remember what got corrupted before those changes? There's a drastic difference between removing the central header for a file and the new implementation which seems to overwrite the rest of the contents of the archive.

As for how I'd want this API to look, I'd currently expect the additions to look like this:

```rs
/// Anonymizes all data of the file currently named `name`
///
/// All of file contents, file name and metadata such as creation date or permission flags are removed from the file.
/// In its place, we leave an empty file. Keep in mind that the bleached file will retain the length of the original.
pub fn bleach_file(&mut self, name: impl Into<String>) -> ZipResult<()>;
/// Removes `name` from the archive's file listing.
///
/// > Warning: The file will remain intact within the archive, and can be recovered by anyone who wants to.
///   This prevents it from being read by a typical ZIP unpackers or viewers.
///   To remove the data contained in the file, use [`ZipWriter::bleach_file`] before this method.
///
/// This can be used to cheaply remove a file from the archive. This does not write to the archive until [`ZipWriter::finalize`] is called.
pub fn forget_file(&mut self, name: impl Into<String>) -> ZipResult<()>;
```

Say if you think I'm missing something here, or you think any of it should be named differently! It's nothing more than a sketch from my memory of how I intended to implement this in november.

baloo:
> Heya! Thankyou for sending this in. As rushii mentioned before, this API isn't appropriate for the public API of the crate :) If the goal is to overwrite the space in a zip archive, recreating the archive from scratch is more suitable.

I understood [@rushiimachine 's comment](https://github.com/zip-rs/zip/issues/283#issuecomment-1314347056) as "I could not implement that with the current public facing API, so I could not use a trait to augment/provide the feature", more than this should not be public facing.
I could be wrong.

baloo:
(anyway, I'll try to put some effort to have the `forget_file` and `bleach_file` the way you suggest)

Plecra:
I'm expecting to implement this myself (lets say by February 19th and put the pressure on ;)), but if you'd like to return to it before then I'm happy to work through it together!

baloo:
I got pulled in side projects, and couldn't get back to that. I don't know if I'll have the time before Feb 19th, but I'd be happy to test your changes if necessary.

Plecra:
That's perfectly reasonable :) Thanks for putting your time into this, and I'll definitely ping you to see how the new implementation suits you!

PR #336, opened 2023-01-20
Add units to the size documentation

jkasari:
The documentation did not list what it meant by "size". I changed it to specify that it is returning the size in bytes.

The standard library fs module specifies that size is returned in bytes. I modified this documentation to match the style of the standard library. 

https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len


zamazan4ik (approved):

zamazan4ik:
Thanks!

PR #337, opened 2023-01-30
Detect insufficient CentralDirectoryEnd record

Mik-pe:
Per zip spec 4.4.1.4 (https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT)

>      4.4.1.4  If one of the fields in the end of central directory
>      record is too small to hold required data, the field SHOULD be 
>      set to -1 (0xFFFF or 0xFFFFFFFF) and the ZIP64 format record 
>      SHOULD be created.

Previously these archives were incorrectly detected as multi-disk.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -348,7 +348,9 @@ impl<R: Read + io::Seek> ZipArchive<R> {
             Some(locator64) => {
                 // If we got here, this is indeed a ZIP64 file.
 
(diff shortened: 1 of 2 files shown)


zamazan4ik (approved):

zamazan4ik:
Thanks!

PR #338, opened 2023-02-01
Fix clippy issues

aquacash5:
This fixes the clippy issues


Plecra (approved):
Oh, thanks a bunch! Will be nice to see all those green ticks again 🙂

Issue #335, opened 2023-01-14
CVE-2023-22895 - bzip2 vulnerability


hampuslidin:
A security advisory has been raised for the bzip2 <0.4.4 versions:

https://www.cvedetails.com/cve/CVE-2023-22895/

Bumping this dependency to 0.4.4 would circumvent the security issue.

Plecra:
Ah, cheers. I'd rather not address this within zip, as dependency versions are really meant for compatibility, not protecting users. I'd prefer the upstream version yanked, but we can bump our version if bzip2 is hesitant to do that.

Plecra:
Ah yanking is still a bit messy. I'll have a look into what the best solution is in zip!

Plecra:
I checked in with the Community Discord, and yeah a dependency bump in zip would be pretty atypical. Do you know which users this change would protect? As I understand it, the only thing we'd manage to do is break some builds which explicitly depend on <0.4.4, which isn't a goal. Every newly built dependent on zip should be using secure versions already.

lnicola:
Yeah, it would be pretty unusual to do this in `zip`.

hampuslidin:
Hi, thanks for the response!Fair enough, since it is possible for dependant users to upgrade bzip explicitly in their Cargo lockfile, it is better to keep compatability with crates that might depend on those earlier versions.You may close this issue in that case!31 jan. 2023 kl. 07:37 skrev Laurențiu Nicola ***@***.***>:﻿
Yeah, it would be pretty unusual to do this in zip.

mcandre:
Backwards compatibility over security patches?

NobodyXu:
> Backwards compatibility over security patches?

For new projects using zip, they would automatically use the latest bzip2 release.
For existing projects using zip, they can always upgrade bzip2 to latest release themselves.

Magicloud:
Why not bump zip version and restrict the bzip2 being used.

lnicola:
This crate doesn't pick a bzip2 version, the binaries do. There is no vulnerability introduced by this crate.

PR #351, opened 2023-02-03
Bump bzip2 version to 0.4.4

bjrjk:
Patched CVE-2023-22895.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2021"
 [dependencies]
 aes = { version = "0.7.5", optional = true }
 byteorder = "1.4.3"
-bzip2 = { version = "0.4.3", optional = true }
+bzip2 = { version = "0.4.4", optional = true }
 constant_time_eq = { version = "0.1.5", optional = true }
 crc32fast = "1.3.2"
 flate2 = { version = "1.0.23", default-features = false, optional = true }


Plecra:
Haha, see #335 . What do you think of the conclusion we came to there?

bjrjk:
Sorry, I wasn't aware of this issue yesterday.

It's reasonable to choose not bumping the version of bzip for compatibility consideration if users can upgrade bzip freely via specifying version in `Cargo.toml` explicitly. However, I think my patch on bzip2 hasn't modified any interface and wouldn't introduce any compatibility problem into downstream crates.

The CVE's problem reproduces when trying to zip or unzip a file large than 4GB. It cause the program stucks. In usual cases, most files are smaller than 4GB and the edge case will seldom appear. But somebody likes me do meet the edge case and they cause my application stucks, so I had to patch it. 🤣

I'm a rust newbie and may not have the capability to weigh up the pros and cons in whether bumping the version. The above is some facts about this vulnerablity. You may freely make your own decision.

Thanks for you replying!

Plecra:
Thankyou for your thoughts! I also don't want users to be unnecessarily exposed to the cve, so it's worth looking at. I'm going to leave the version alone for two reasons:

- any newly code will be compiling with 0.4.4 already, and be secure against the vulnerability. This is cargo's default behaviour
- raising the required version unfortunately *is* a breaking change. This is a bit subtle: if users have a project building already with `zip` and also another dependency clamping bzip2's version, cargo won't be able to resolve the dependency tree and the build will error out. This is fairly unique to our case: it's only a problem because bzip2 uses c code underneath

PR #354, opened 2023-02-12
update aes to 0.8

### LINKED ISSUE #352: Updating aes crate.
We have been looking into upgrading the "aes" crate in debian (it won't happen now until after the bookworm freeze), and one of the reverse dependencies is the zip crate. While we do have a mechanism for packaging multiple versions of crates it is something we try to avoid where possible.

Is there any chance of an update of the zip crate to use the latest version of the aes crate. I tried bumping the dependency but quickly ran into build errors.

Plecra:
(hazard): The changelog in aes doesn't indicate any change in behaviour, and I have assumed the new APIs work identically

This closes #352. Lets check that the CI likes it.

@mbr! As the author of the original PR, would love your input here - do you think any more changes need to be made to support the newer `aes`?

### DIFF
--- a/src/aes_ctr.rs
+++ b/src/aes_ctr.rs
@@ -2,10 +2,11 @@
 //!
 //! This was implemented since the zip specification requires the mode to not use a nonce and uses a
 //! different byte order (little endian) than NIST (big endian).
-//! See [AesCtrZipKeyStream](./struct.AesCtrZipKeyStream.html) for more information.
+//! See [AesCtrZipKeyStream] for more information.
 
 use aes::cipher::generic_array::GenericArray;
-use aes::{BlockEncrypt, NewBlockCipher};
+// use aes::{BlockEncrypt, NewBlockCipher};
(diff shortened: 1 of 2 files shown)


Plecra:
I'll merge this on Tuesday if we don't find a reason it shouldn't be added :)

src/aes_ctr.rs
```diff
@@ -2,10 +2,11 @@
 //!
 //! This was implemented since the zip specification requires the mode to not use a nonce and uses a
 //! different byte order (little endian) than NIST (big endian).
-//! See [AesCtrZipKeyStream](./struct.AesCtrZipKeyStream.html) for more information.
+//! See [AesCtrZipKeyStream] for more information.
 
 use aes::cipher::generic_array::GenericArray;
-use aes::{BlockEncrypt, NewBlockCipher};
+// use aes::{BlockEncrypt, NewBlockCipher};
```
aquacash5, line 8: Should remove unused commented code.

PR #360, opened 2023-03-29
Add rust-version field to Cargo.toml

jturner314:
This causes compilers older than the MSRV to produce a warning/error which more clearly indicates that the crate needs a newer compiler. See https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ description = """
 Library to support the reading and writing of zip files.
 """
 edition = "2021"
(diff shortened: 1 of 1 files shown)


Plecra:
fantastic, thankyou

Plecra:
Manually merged in https://github.com/zip-rs/zip/commit/13a19f682f8d306dbb006db70d97daf6b4446776

PR #363, opened 2023-04-11
Add AES encryption write support

kyrias:
This pull request adds support for writing AES encrypted ZIP files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,10 @@ crc32fast = "1.3.2"
 flate2 = { version = "1.0.23", default-features = false, optional = true }
 hmac = { version = "0.12.1", optional = true, features = ["reset"] }
 pbkdf2 = {version = "0.11.0", optional = true }
+rand = { version = "0.8.5", optional = true }
 sha1 = {version = "0.10.1", optional = true }
 time = { version = "0.3.7", optional = true, default-features = false, features = ["std"] }
+zeroize = { version = "1.6.0", optional = true, features = ["zeroize_derive"] }
 zstd = { version = "0.11.2", optional = true }
 
 [target.'cfg(any(all(target_arch = "arm", target_pointer_width = "32"), target_arch = "mips", target_arch = "powerpc"))'.dependencies]
@@ -34,7 +36,7 @@ walkdir = "2.3.2"
 time = { version = "0.3.7", features = ["formatting", "macros"] }
 
 [features]
-aes-crypto = [ "aes", "constant_time_eq", "hmac", "pbkdf2", "sha1" ]
+aes-crypto = ["aes", "constant_time_eq", "hmac", "pbkdf2", "sha1", "rand", "zeroize"]
 deflate = ["flate2/rust_backend"]
 deflate-miniz = ["flate2/default"]
 deflate-zlib = ["flate2/zlib"]
--- a/examples/write_sample.rs
+++ b/examples/write_sample.rs
@@ -1,5 +1,5 @@
 use std::io::prelude::*;
-use zip::write::FileOptions;
+use zip::{write::FileOptions, AesMode};
 
 fn main() {
     std::process::exit(real_main());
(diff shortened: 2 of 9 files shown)


kyrias:
Ping.  :)

svenstaro:
@Plecra Hey, I'd really like to see this go in. Could you please take a look?

kyrias:
One thing I just realized is that there is either something wrong when we encrypt a file where the real compression mode is stored, or 7z's implementation is wrong, because `7z t` complains about an incorrect CRC and then fails to decrypt the following file, implying that there's some kind of offset issue.

Plecra:
Hey! I haven't figured out how to responsibly review a PR for AES support. Since it could potentially be used for real data protection, I wouldn't want to rely on my patchy understanding.

Do you have any recommendations for how I should approach this PR? (naturally, since I can't judge a good implementation, I also can't judge your expertise :P)

kyrias:
> One thing I just realized is that there is either something wrong when we encrypt a file where the real compression mode is stored, or 7z's implementation is wrong, because `7z t` complains about an incorrect CRC and then fails to decrypt the following file, implying that there's some kind of offset issue.

Figured it out, the authentication code was never being written out.



> Hey! I haven't figured out how to responsibly review a PR for AES support. Since it could potentially be used for real data protection, I wouldn't want to rely on my patchy understanding.
> 
> Do you have any recommendations for how I should approach this PR? (naturally, since I can't judge a good implementation, I also can't judge your expertise :P)
> 
> (edit: also, if you think this is a false premise, lmk how :))

Did you see the reply I sent to your message on Mastodon?

Plecra:
Hm... no I didn't :D should it be here?
image

Plecra:
nada 
image

Plecra:
Could the mastodon instance I'm on be a problem? Anyway, do you want to just send the reply here :)

kyrias:
So for starters I can say that since I haven't introduced any new crypto code there isn't that much to watch out for.  The encryption code paths uses the exact same crypto code that is already being used for decrypt, and those are in the end using the RustCrypto `aes` crate.

I would recommend reading the actual [WinZip AES encryption specification](https://www.winzip.com/en/support/aes-encryption/) as well, since it's fairly short and readable.

These are the three main potential concerns I can come up with given the above.


# Salt

The salt that is used together with the password to derive the actual encryption keys need to be generated in a cryptographically secure manner.  If you accidentally encrypt two different files with the same password and salt the two files end up leaking information about each-other.

To generate the salt we're using the `ThreadRng` from the `rand` crate.  This is backed by a cryptographically secure PRNG, and so we should be safe here.


# CRC-32

Another thing to watch out for is not to use the AE-1 version of the encryption specification with files shorter than 20 bytes.  For files that are that short the CRC-32 would allow recovering the plaintext.  Therefore we follow their current recommendations of using AE-2 for files shorter than 20 bytes, which really just means that we zero out the CRC-32 field of the file, and otherwise we use AE-1 which sets the CRC-32 as usual.


# Leaving things in memory

A third thing that some people might mind is the possibility of non-zeroized secret data being left in memory, which could potentially be recovered.

This is something that is already not really handled by the crate already and if wanted is something that should be done more in-depth, but I did at least make sure that the buffers used for the un/encrypted file and for the derived encryption key are both zeroized.

kyrias:
Ah, it's going to take a bit to rework this on top of the new ZipCrypto support.

kyrias:
I figured out the rework necessary, so please review when you have the time.  :)

kbknapp:
> **Warning**
> This isn't a cryptographic review; only a functional anecdote

I've been using this branch in a POC project and everything appears to work as desired.

kyrias:
Friendly ping.  :)

svenstaro:
@Plecra Hey, could we get this merged? As we see, people are using this successfully. I think it'd be better than to have people run other people's branches if they want zip security. :)

Plecra:
I appreciate the bump, and I think you're right. I'll try to get to this asap.

svenstaro:
@Plecra Bumpybump :)

svenstaro:
Sorry @Plecra I hate to do this again but is there any chance you could look at this? We've been using this in production for some time now and we think it's good. Would be great to just get this mainlined.

nickparser:
hey, would you mind reviewing and merging this pull request? :)

Pr0methean:
This repo is no longer maintained. Could you please rebase this PR against https://github.com/zip-rs/zip2/ and reopen it there?

PR #361, opened 2023-04-02
Add support for deflate-zlib-ng feature

JakubOnderka:
zlib-ng is faster version of zlib library

See https://github.com/zlib-ng/zlib-ng

PR #365, opened 2023-04-29
Add deflate-zlib-ng feature flag

jirutka:


Plecra:
haha yeah, I've never figured out best practices for dependencies' features. Do you think there's a reason we should forward this feature but not every other feature of all our dependencies? (I know that crates will often instead depend directly on `flate2` and specify the features themselves)

Though I realize we're already doing this with the other `deflate` backends. You'll need to add `deflate-zlib-ng` to the configuration gates, otherwise the crate wont enable support properly. We need tests to pass with `cargo test --no-default-features --features deflate-zlib-ng`

JakubOnderka:
I already send this PR: https://github.com/zip-rs/zip/pull/361 that's also adds the configuration gates. And I will also run tests with this feature flag.

Pr0methean:
This was done in https://github.com/zip-rs/zip2/commit/9ec0ddc5ca09bb2a73374941bef48615267297fb.

PR #368, opened 2023-05-10
Added function to get if a file is encrypted or not

reduf:
Hi,

the reason why this is useful is that file can be encrypted or not and sometime you may only want to ask for the password for files that are encrypted, but work without for non-encrypted files. Being able to know if the file is encrypted is very useful for that.


Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/64.

Issue #223, opened 2021-02-14
Seek trait support for RawReader/Stored files


zumoshi:
As per #221, @Plecra said ZipFile itself won't receive seek trait but seek support is possible through RawReader or a checked version that could provide another type of reader for stored files (no encryption/compression) with seek.
Additionally, a method that just returns bounds (a tuple with first/last byte index of file) could be useful.

I had some issues with passing ZipFile through an UnwindSafe barrier. but a method that just returns a tuple of (u64,u64) for (start, end) of a zipfile inside archive can easily be passed then a normal file with limit could be used to read that part.

Thanks for your time.

busstoptaktik:
That would definitely be very useful - I'm hoping to be able to use a zip archive as container format for large, files, kept uncompressed ("stored"), for direct, read-only access to individual bytes on disk.

So searching the zip-file by stored file's name and getting a tuple of `(usize, usize)` in return would be the key for me.

exFalso:
Just noting that I'm picking this up in a fork as we need it. We also use zip without compression/encryption. Will submit a PR

Pr0methean:
This may be doable by translating https://github.com/madler/zlib/blob/develop/examples/zran.c to Rust.

Pr0methean:
This repo is no longer maintained. This is being addressed in the case of Stored files by https://github.com/zip-rs/zip2/pull/69.

PR #369, opened 2023-05-10
Add by_name_seek() for Stored zips

exFalso:
This PR adds a `ZipArchive::by_name_seek` function which returns a `ZipFileSeek` structure that implements `Seek`.
This only works for `Stored` types of zips for now, although I'm sure this can be implemented for compressed archives as well using some kind of lazy indexing structure that helps map compressed indexes to uncompressed ones quickly.

I tried various ways of adding this feature, but I could only make it work like this. I "forked" the `ZipFile` structure, added a `ZipFileSeek`, and added a common trait that implements the metadata goodies. The reason I couldn't make it work with a unified approach is because the current library supports non-`Seek` base readers, and there's no such thing as a dynamic cast in Rust (nor would I want there to be).

One consequence of the trait is that this change became **API breaking**, as the trait must be imported to access the metadata functions. I can add back the original functions that dispatches to the trait in case this is an issue.

Although I have not added tests to this crate, it has been tested fairly thoroughly with tests testing our own usage of the feature.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,6 +26,7 @@
 
 pub use crate::compression::{CompressionMethod, SUPPORTED_COMPRESSION_METHODS};
 pub use crate::read::ZipArchive;
+pub use crate::read::HasZipMetadata;
 pub use crate::types::DateTime;
 pub use crate::write::ZipWriter;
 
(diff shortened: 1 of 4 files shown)


exFalso:
Linking with issue https://github.com/zip-rs/zip/issues/223

Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/69.

PR #370, opened 2023-05-13
Fixed writing wrong UTF-8 flag

hidez8891:
This PR is related to #239.

If the UTF-8 flag (general purpose bit 11) is set, file names and comments MUST be saved in UTF-8. (APPENDIX D)
https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

However, the UTF-8 flag is set even for formats that are non-UTF-8 (GB18030, SHIFT_JIS, etc).


Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/62.

PR #373, opened 2023-06-09
Add ZipWriter::set_file_metadata()

exFalso:
This PR adds a method to `ZipWriter` which overwrites the internal crc32 calculation.

Use case: in our specific scenario we need to write to the inner `Write` outside of `ZipWriter` (tldr: we deal with very large files which need to be streamed/managed independently), but we need `ZipWriter` to be aware of this in order to write the correct length and crc32 values to the zip metadata.

### DIFF
--- a/src/write.rs
+++ b/src/write.rs
@@ -329,6 +329,22 @@ impl<W: Write + io::Seek> ZipWriter<W> {
         self.comment = comment;
     }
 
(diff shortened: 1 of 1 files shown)


Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/66.

PR #374, opened 2023-06-12
Update Zstd to 1.5.2

startewho:
https://github.com/gyscos/zstd-rs has update a lot since 0.11.
We should flowing the source lib

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ hmac = { version = "0.12.1", optional = true, features = ["reset"] }
 pbkdf2 = {version = "0.11.0", optional = true }
 sha1 = {version = "0.10.1", optional = true }
 time = { version = "0.3.7", optional = true, default-features = false, features = ["std"] }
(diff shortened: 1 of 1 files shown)


a1phyr:
Version `0.13` of `zstd` is out

Cargo.toml
```diff
@@ -22,7 +22,7 @@ hmac = { version = "0.12.1", optional = true, features = ["reset"] }
 pbkdf2 = {version = "0.11.0", optional = true }
 sha1 = {version = "0.10.1", optional = true }
 time = { version = "0.3.7", optional = true, default-features = false, features = ["std"] }
-zstd = { version = "0.11.2", optional = true }
+zstd = { version = "0.12", optional = true }
```
vijay-10x, line 25:
```suggestion
zstd = { version = "0.13", optional = true }
```

Pr0methean:
This repo is no longer maintained, and https://github.com/zip-rs/zip2 already uses zstd 0.13.1.

PR #383, opened 2023-07-23
Implement ZipWriter::new_streaming for unseekable writers

axnsan12:
This is a revival of #114 to implement unseekable writing via data descriptors. 

No public API is changed except for the addition of `ZipWriter::new_streaming`, and all adjacent functionality still works - zip64, extra data, ZipCrypto encryption. I even fixed a bug where encryption was broken if used together with extra data :).


chenxiaolong:
Thanks a lot for implementing this!

I've been testing it quite a bit and it's been working very well. I only found a single issue where a panic occurs with the streaming ZipWriter when a non-zip64 small file entry is written such that the local header offset is >=2^32. I fixed it with: https://github.com/chenxiaolong/zip/commit/baf807a5bedbb068a8d48189be6223decfca5f37.

Pr0methean:
This repo is no longer maintained. Could you please rebase this against https://github.com/zip-rs/zip2 and reopen it there?

Issue #382, opened 2023-07-22
zip file created on windows 11 can't be unzipped


xudesheng:
My customer shared a zip file (too big to attach here), which he created from his Windows 11 machine with the built-in function.

I can't double-click on my Mac to unzip it (I believe it's using the `Archive Utility`), but I can use the "unzip" command to unzip it.

when I use this crate to unzip it, the crate reports an error: `unsupported Zip archive: Compression method not supported.`

any idea how to export the compression method in the error message?

xudesheng:
Unfortunately, it's due to:`pub const DEFLATE64: Self = CompressionMethod::Unsupported(9);`

xudesheng:
This issue is about how to support `deflate64` when unzip.
I submitted an issue to the `flate2-rs` crate at: https://github.com/rust-lang/flate2-rs/issues/365
notice this new crate pops up: https://crates.io/crates/deflate64

hope zip-rs can leverage it to support `deflate64` when unzip.

Pr0methean:
Fixed by https://github.com/zip-rs/zip2/pull/18.

PR #384, opened 2023-07-23
a new example to show how to write large files (greater than 4GB)

Byron:
This requires DEFLATE64 support, which is seemingly present in the default rust backend.

This example is motivated by a possibly [related issue](https://github.com/rust-lang/flate2-rs/issues/365) over at `flate2`.

### DIFF
--- a/examples/write-large-file.rs
+++ b/examples/write-large-file.rs
@@ -1,0 +1,36 @@
+//! Write a huge file with lots of zeros, that should compress perfectly.
+
+use std::io::Write;
+
+use zip::write::FileOptions;
+fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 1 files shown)


axnsan12:
I believe the flate2 issue OP is referring to decompression of deflate64 (see #382 here)

xudesheng:
> I believe the flate2 issue OP is referring to decompression of deflate64 (see #382 here)

@axnsan12 Thank you, yes, my original issue is how to decompress the file compressed with the `deflate64` algorithm.

@Byron Thank you anyway for your example. It's useful. My current focus is on how to decompress

Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/67.

PR #389, opened 2023-08-14
support more 32bit target

Dirreke:
support more 32bit target like csky


Pr0methean:
This was rendered obsolete by https://github.com/zip-rs/zip2/commit/9438bef0f7577cd7babd3d9bd52d575e71109e9b.

PR #390, opened 2023-08-18
Remove byteorder dependency

psychon:
Hi,

I came across the use of "byteorder" in this crate and thought "one does not really need byteorder for this". Feel free to disagree and to just close this PR.

Since there are many uses of "byteorder", I tried to do this in small parts and do a commit for each one. Hopefully this makes it easier to understand what is going on. Dunno.

Then I noticed that actions are failing in my fork. The last few commits make rustfmt and clippy happy. These changes are unrelated to this PR. Feel free to cherry pick what you want.

Note that CI will still fail, because:
```
error: package `base64ct v1.6.0` cannot be built because it requires rustc 1.60 or newer, while the currently active rustc version is 1.59.0
```
I'll leave that one to you.


Pr0methean:
This repo is no longer maintained. Could you please rebase this change against https://github.com/zip-rs/zip2 and submit it there?

PR #393, opened 2023-08-29
fix: clippy lints and formatting (rustfmt)

striezel:
This pull request fixes the current warnings issued by `clippy`.


striezel:
```
error: package `base64ct v1.6.0` cannot be built because it requires rustc 1.60 or newer, while the currently active rustc version is 1.59.0
```

The MSRV builds are failing, because somewhere in there `base64ct` v1.6.0 is pulled in (via dependencies `pbkdf2` -> `password-hash`) and v1.6.0 has bumped the MSRV to 1.60: https://github.com/RustCrypto/formats/blob/master/base64ct/CHANGELOG.md
So it's probably time to bump the MSRV of `zip`, too, but I leave that for another PR.

Pr0methean:
This repo is no longer maintained, and https://github.com/zip-rs/zip2 already passes Clippy and `cargo fmt --all -- --check`.

PR #394, opened 2023-09-07
use cfg-if and cfg_attr() to improve readability of feature guards

cosmicexplorer:
1. The [`cfg-if`](https://docs.rs/cfg-if/latest/cfg_if/) crate makes it easier to read conditional compilation logic.
2. There is an undocumented hack used by many crates to get `docs.rs` builds to show whether some value is only available with a given feature enabled (see e.g. rust-random/rand#986).
    - The result can be tested with `RUSTDOCFLAGS='--cfg docsrs' cargo doc` (try opening e.g. `CompressionMethod` to see how this shows which enum cases are supported by which features).

This is not a breaking change, as `cfg-if` should generate the same code as before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,7 @@ harness = false
 [[bench]]
 name = "read_metadata"
 harness = false
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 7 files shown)


Pr0methean:
The `cfg-if` crate was made redundant by the [cfg! macro](https://doc.rust-lang.org/std/macro.cfg.html), which makes it possible to check configuration flags in a normal `if` statement. Even if that weren't the case, it doesn't look to me like this PR adds any readability -- the `cfg-if!` is just another pair of curlies to keep track of, and the `cfg!` macros remain just as complicated.

This repo is unmaintained. If you'd still like to submit an improvement, do so at https://github.com/zip-rs/zip2.

PR #395, opened 2023-09-07
use num_enum to clean up the System type

cosmicexplorer:
The [`num_enum`](https://docs.rs/num_enum/latest/num_enum/) crate will automatically derive `From` implementations for `enum`s with `repr(u8)` or similar. This change removes a small amount of boilerplate.

This is not a breaking change, because `System` is not exported by this crate.


Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/63.

PR #396, opened 2023-09-07
use indexmap in read::zip_archive::Shared instead of a separate vec and hashmap

cosmicexplorer:
The [`indexmap`](https://docs.rs/indexmap/latest/indexmap/index.html) crate offers a hashmap which tracks the order each element was added. This lets us remove some boilerplate around accessing zip file entries by name or by index.

This is not a breaking change, but may improve performance and reduce memory usage.


Pr0methean:
Because this repo is unmaintained, I've replaced this PR with https://github.com/zip-rs/zip2/pull/50.

PR #397, opened 2023-09-07
use displaydoc and thiserror to remove some boilerplate

cosmicexplorer:
The [`displaydoc`](https://docs.rs/displaydoc/latest/displaydoc/) and [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) crates automatically derive much of what is written in `result.rs`. This change removes some boilerplate.

This is not a breaking change, because it should generate the same code as before.


Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/65.

PR #398, opened 2023-09-07
merge InvalidPassword into ZipError

cosmicexplorer:
A lot of the logic in `read.rs` goes through functions that return a nested `Result<Result<...>, InvalidPassword>`, which seems unnecessary when we don't even have password support at all in some configurations. This change makes `InvalidPassword` into a case of `ZipError` and avoids needing to separately unwrap an `InvalidPassword` error.

**This is a breaking change, because it modifies `zip::result::ZipError` and removes `zip::result::InvalidPassword`.**


Pr0methean:
Rendered obsolete by https://github.com/zip-rs/zip2/commit/ece098d39332f97b04e508222d0cc44ae3c1e5c4.

PR #399, opened 2023-09-07
provide error cases for DateTime parsing

cosmicexplorer:
1. Currently, there is no indication of why a `DateTime` construction failed. Since zip files have a strange representation for dates and times, it's helpful to specify why something failed.
    - This change adds multiple cases to an enum `DateTimeRangeError`.
    - This change deprecates `DateTime::from_date_and_time()`, which ignores `DateTimeRangeError`, and adds `DateTime::parse_from_date_and_time()`, which retains the error case.
2. The `DateTime::to_time()` method currently assumes the desired UTC offset is GMT+-0 (GMT). We would like to make this configurable, so that if a zip file was created by a machine with a different UTC offset, we can invert that when reading the zip file.
    - This change deprecates `DateTime::to_time()` and adds `DateTime::to_time_with_offset()`.
3. It's not obvious what value is returned by `DateTime::default()`.
    - This change adds a new `const` method `DateTime::zero()` to make the use of the minimum possible date/time explicit.

This is not a breaking change, since `DateTimeRangeError` was previously an empty struct, and no method signatures are changed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,8 @@ byteorder = "1.4.3"
 bzip2 = { version = "0.4.3", optional = true }
 constant_time_eq = { version = "0.1.5", optional = true }
 crc32fast = "1.3.2"
+displaydoc = "0.2.4"
+thiserror = "1.0.48"
 flate2 = { version = "1.0.23", default-features = false, optional = true }
 hmac = { version = "0.12.1", optional = true, features = ["reset"] }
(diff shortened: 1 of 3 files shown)


Pr0methean:
This repo is no longer maintained; could you please reopen this as a PR against https://github.com/zip-rs/zip2/? Also, the error struct doesn't need to indicate the expected range, since it'll always be the same for a given time unit.

PR #400, opened 2023-09-07
add ZipWriter::finish_into_readable()

cosmicexplorer:
In my project [`medusa-zip`](https://github.com/cosmicexplorer/medusa-zip/), I often write to a zip file and then immediately read it back. Writing a zip file somewhere and then providing a `ZipArchive` view over its contents seems like a reasonable enough use case to support upstream.

This is not a breaking change.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -291,13 +291,41 @@ fn make_reader(
     }
 }
 
+impl<R> ZipArchive<R> {
+    pub(crate) fn from_finalized_writer(
+        files: Vec<ZipFileData>,
+        comment: Vec<u8>,
+        reader: R,
+    ) -> ZipResult<Self> {
+        if files.is_empty() {
+            return Err(ZipError::InvalidArchive(
+                "attempt to finalize empty zip writer into readable",
+            ));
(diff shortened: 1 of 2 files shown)


src/read.rs
```diff
@@ -399,18 +427,18 @@ impl<R: Read + io::Seek> ZipArchive<R> {
     ///
     /// This uses the central directory record of the ZIP file, and ignores local file headers
     pub fn new(mut reader: R) -> ZipResult<ZipArchive<R>> {
-        let (footer, cde_start_pos) = spec::CentralDirectoryEnd::find_and_parse(&mut reader)?;
+        let (footer, cde_end_pos) = spec::CentralDirectoryEnd::find_and_parse(&mut reader)?;
```
Pr0methean, line 430: Revert this name change. CDE stands for Central Directory End, and cde_start_pos is the start of the CDE record.

Pr0methean (changes requested):

Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/57.

PR #402, opened 2023-09-07
publicly export and document the zip64 threshold constants

cosmicexplorer:
Currently, there's no way for me to know whether I should set `FileOptions::large_file()` to `true` for a given zip file entry, because I don't know how large the file needs to be to require that. This change publicly exposes `ZIP64_BYTES_THR` and `ZIP64_ENTRY_THR` along with useful docstrings to enable that use case.


Pr0methean:
Replaced by https://github.com/zip-rs/zip2/pull/60.

PR #405, opened 2023-09-10
Bump constant_time_eq to 0.3

bowlofeggs:
0.3.0 is the latest release. The primary motivator for this change is to eliminate the CC0 license from the dependency tree, which was fixed in constant_time_eq 0.2.4:

  https://github.com/cesarb/constant_time_eq/commit/0738c1a579343d31406d59ed9354f2f92e896bce

This change also raises the minimum supported Rust version to 1.66 since that is the MSRV for constant_time_eq.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,13 +9,13 @@ description = """
 Library to support the reading and writing of zip files.
 """
 edition = "2021"
-rust-version = "1.59.0"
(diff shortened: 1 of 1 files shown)


Pr0methean:
This repo is no longer maintained, and https://github.com/zip-rs/zip2/ is already using constant_time_eq 0.3.0.

PR #408, opened 2023-09-22
parameterize ZipFile by stream type and remove the lifetime so it can be Send + 'static

cosmicexplorer:
It's currently difficult to use this crate in `async` Rust code. One reason for that is the requirement that all captured variables in closures for most async executors must be `Send + 'static`, neither of which is true for `ZipFile<'a>` right now, requiring `unsafe` workarounds for the lifetime rules (e.g. casting a pointer to a usize, as in https://github.com/cosmicexplorer/symbolic-fs/blob/827a71e72cd7e3f5b7fd3fe2a0f3ad836508eff6/handles/src/lib.rs#L479).

This change does two things:
1. Parameterize `ZipFile<S>` by the type of the inner readable stream `S`, and remove the lifetime requirement.
2. Roll our own `Limiter<S>` to perform the role of `io::Take<&mut dyn R>`.

This is unfortunately a breaking change because `ZipFile` is a public API.

### DIFF
--- a/src/aes.rs
+++ b/src/aes.rs
@@ -113,7 +113,7 @@ impl<R: Read> AesReader<R> {
 /// There is a 1 in 65536 chance that an invalid password passes that check.
 /// After the data has been read and decrypted an HMAC will be checked and provide a final means
 /// to check if either the password is invalid or if the data has been changed.
-pub struct AesReaderValid<R: Read> {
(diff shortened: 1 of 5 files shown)


Pr0methean:
This repo is no longer maintained. Could you please rebase this PR against https://github.com/zip-rs/zip2 and submit it there?

Issue #413, opened 2023-10-26
Use cfg(target_has_atomic) instead of hard coding which targets support atomics


bjorn3:
This has been stable since 1.60.0: https://github.com/rust-lang/rust/pull/93824 The hard coded list is bound to go out of sync with rustc, especially when new targets get added. In fact it is already wildly out of sync. You currently only disable 64-bit atomics for 32bit arm, powerpc and mips, but a lot more targets like riscv32, sparc and thumb also don't support them: https://github.com/crossbeam-rs/crossbeam/blob/5fd21b032388c2e418c7306ba6149e1881943c31/no_atomic.rs#L19-L78 As rustc itself depends on crossbeam, this means it isn't possible to build rustc for targets lacking some atomics until a new crossbeam release with those new targets. Using cfg(target_has_atomic) will immediately work.

bjorn3:
cc https://github.com/rust-lang/rust/pull/117170#discussion_r1373230638

Pr0methean:
This was rendered obsolete by https://github.com/zip-rs/zip2/commit/9438bef0f7577cd7babd3d9bd52d575e71109e9b.

PR #422, opened 2023-12-27
Added xtensa architecture to cfg for AtomicU64

gpgreen:
I am using the epub crate, which has zip as a dependencie for a xtensa project. It won't compile as this architecture doesn't have 64bit atomics


kyrias:
(Doing this would then close #413.)

Pr0methean:
This was rendered obsolete by https://github.com/zip-rs/zip2/commit/9438bef0f7577cd7babd3d9bd52d575e71109e9b.

PR #426, opened 2024-01-09
Use clap to parse args and add compression method argument in write_dir example

allilo1:
In the write_dir example, the compression methods are iterated over in a loop and the resulting output zip file is repeatedly overwritten. Since ZStd is the last compression method in the loop, the output zip file is compressed using ZStd. 

This change updates the write_dir example to use clap and adds the compression method as an argument. Deflated-Miniz and Deflated-Zlib are broken out as separate methods based on the specified feature. If the feature needed for the compression method is not specified, we exit from main with an error.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ rust-version = "1.59.0"
 aes = { version = "0.8.2", optional = true }
 byteorder = "1.4.3"
 bzip2 = { version = "0.4.3", optional = true }
-clap = { version = "4.4.11", features = ["derive"] }
 constant_time_eq = { version = "0.1.5", optional = true }
 crc32fast = "1.3.2"
 flate2 = { version = "1.0.23", default-features = false, optional = true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -15,6 +15,7 @@ rust-version = "1.59.0"
 aes = { version = "0.8.2", optional = true }
 byteorder = "1.4.3"
 bzip2 = { version = "0.4.3", optional = true }
+clap = { version = "4.4.11", features = ["derive"] }
```
DrChat, line 18: You should add this as a [`dev-dependency`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies) instead (as you're only using `clap` for an example).

allilo1: Thank you for pointing this out, I moved `clap` to `dev-dependencies`

Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/59.

PR #429, opened 2024-01-26
Add rs versions of bzip and zstd in order to pass static compiles

lolpro11:
Bzip-sys and Zstd-sys are the only two packages that require dynamic linking to system libs.
In order to pass a static compile (for use in containers), I recommend using a pure Rust impl of these compression algorithms.
Zstud seems to be using ffi functions, but it does produce a static lib. It does use system libs for compiling, which should be fine.


Pr0methean:
Replaced by https://github.com/zip-rs/zip2/pull/58.

PR #433, opened 2024-02-21
do not panic in Drop when stream is in an error state

### LINKED ISSUE #432: panic when trying to exhaust the stream
The Drop implementation of `ZipFile` currently panic's when the stream has an issue. Unfortunately we hit this panic sometimes when passing in a reqwest stream into the zipfile (using the `read_zipfile_from_stream` function).

Would it not be better to ignore errors?

https://github.com/zip-rs/zip/blob/3e88fe66c941d411cff5cf49778ba08c2ed93801/src/read.rs#L1008

Any further operations would anyways error with this particular stream...

wolfv:
fixes #432 

This causes a panic for us when a stream errors (we pass in a reqwest stream): https://github.com/prefix-dev/pixi/issues/799

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -1004,8 +1004,10 @@ impl<'a> Drop for ZipFile<'a> {
                 match reader.read(&mut buffer) {
                     Ok(0) => break,
                     Ok(_) => (),
-                    Err(e) => {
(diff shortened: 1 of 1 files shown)


Pr0methean:
Because this repo is no longer maintained, I've replaced this PR with https://github.com/zip-rs/zip2/pull/48.

PR #434, opened 2024-02-26
Add remove and update file examples

jmacadie:
Per [discussion #430](https://github.com/zip-rs/zip/discussions/430), it was not obvious to me how to update a file within an archive. I have added some examples of how to delete and update to make it easier for anyone in a similar situation to me in the future


Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/55.

PR #436, opened 2024-03-02
Restore original filetime when unzip

mass10:
## Summary
 Want original file timestamp be restored when unzip-ing.

## Remark
* Using chrono


Pr0methean:
This repo is no longer maintained, so I've replaced this PR with https://github.com/zip-rs/zip2/pull/46.

mass10:
@Pr0methean Thank you!

PR #441, opened 2024-04-05
Improve several `Read` methods on `ZipFile`

a1phyr:
This notably optimizes the common `zip_file.read_to_end(&mut buf)` case.


Pr0methean:
This repo is no longer maintained, so I've replaced this PR with https://github.com/zip-rs/zip2/pull/45.

Issue #440, opened 2024-04-04
Append to existing archive


TheOnAndOnlyZenomat:
Hello,
I am trying to append to an already existing archive with files. Browsing the docs and the repo I found `new_append()` and according to the PR #215 it should update the central directory header and make the new files available next to the already existing files in the archive.

```rust
use std::{
    fs::{File, OpenOptions},
    io::{Read, Write},
    path::{Path, PathBuf},
    str::FromStr,
};
use zip;

fn gather_files<'a, T: Into<&'a Path>>(path: T, files: &mut Vec<PathBuf>) {
    let path: &Path = path.into();

    for entry in path.read_dir().unwrap() {
        match entry {
            Ok(e) => {
                if e.path().is_dir() {
                    gather_files(e.path().as_ref(), files);
                } else if e.path().is_file() {
                    files.push(e.path());
                }
            }
            Err(_) => todo!(),
        }
    }
}

fn main() {
    let archive = PathBuf::from_str("./archive.zip").unwrap();
    let to_archive = PathBuf::from_str("./asdf/").unwrap();

    let existing_zip = OpenOptions::new()
        .read(true)
        .write(true)
        .append(true)
        .open(&archive)
        .unwrap();
    let mut append_zip = zip::ZipWriter::new_append(existing_zip).unwrap();

    let mut files: Vec<PathBuf> = vec![];
    gather_files(to_archive.as_ref(), &mut files);

    let mut buf: Vec<u8> = vec![];
    for file in files {
        append_zip
            .start_file(file.to_string_lossy(), Default::default())
            .unwrap();

        let mut f = File::open(file).unwrap();
        f.read_to_end(&mut buf).unwrap();

        append_zip.write_all(&buf).unwrap();
    }

    append_zip.finish().unwrap();
}
```

Above is my test program. It should append the files from `./asdf/` to `./archive.zip`. Executing it the first time works and doesn't crash. But when inspecting the archive with `7z l ./archive.zip` I get a warning `WARNINGS: There are data after the end of archive` and the new files don't appear in the file listing. Running the above test program again it crashes with `InvalidArchive("Invalid Central Directory header")` on `new_append(existing_zip).unwrap()`.

Am I using this API wrong, did I miss it's intented usage or is this a genuine bug. If the latter is the case, I'd be very happy to try and fix it (maybe with some guidance) or provide you with testcasts/do profiling and testing.

a1phyr:
[...]

PR #443, opened 2024-04-10
add append example

TheOnAndOnlyZenomat:
Because of my previous confusion in #440, I have written an example to help other avoid that confusion and to make it easier to find this functionality.

### DIFF
--- a/examples/append.rs
+++ b/examples/append.rs
@@ -1,0 +1,68 @@
+use std::{
+    fs::{File, OpenOptions},
+    io::{Read, Write},
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+use zip;
+
+fn gather_files<'a, T: Into<&'a Path>>(path: T, files: &mut Vec<PathBuf>) {
+    let path: &Path = path.into();
+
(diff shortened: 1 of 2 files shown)


examples/append.rs
```diff
@@ -0,0 +1,68 @@
...
+    for file in files {
+        append_zip
+            .start_file(file.to_string_lossy(), Default::default())
+            .unwrap();
+
+        let mut f = File::open(file).unwrap();
+        f.read_to_end(&mut buf).unwrap();
+
+        append_zip.write_all(&buf).unwrap();
+        buf.clear();
```
a1phyr, line 58: I think that it would be better to use `std::io::copy` here

TheOnAndOnlyZenomat: I pushed a commit. I have a suspicion, but could you elaborate as to why `io::copy` is better suited?

TheOnAndOnlyZenomat: Yikes, yeah, that makes indeed sense

Pr0methean:
Replaced with https://github.com/zip-rs/zip2/pull/55.

PR #444, opened 2024-04-15
Matrix alias

Pr0methean:
Fixes a bug where pull-request checks against previous MSRVs are still listed as "expected" if the MSRV is increased while the PR is open. Also eliminates the need to update the ruleset when bumping the MSRV.


Pr0methean:
Meant to send this to the mainline of the source repo.

PR #452, opened 2024-05-04
Use clap to parse args and add compression method argument in write_dir example

allilo1:
In the write_dir example, the compression methods are iterated over in a loop and the resulting output zip file is repeatedly overwritten. Since ZStd is the last compression method in the loop, the output zip file is compressed using ZStd.

This change updates the write_dir example to use clap and adds the compression method as an argument. Deflated-Miniz and Deflated-Zlib are broken out as separate methods based on the specified feature. If the feature needed for the compression method is not specified, we exit from main with an error.

Picked up changes from [zip-rs#426](https://github.com/zip-rs/zip-old/pull/426), and pinned the clap version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ bencher = "0.1.5"
 getrandom = "0.2.5"
 walkdir = "2.3.2"
 time = { version = "0.3.7", features = ["formatting", "macros"] }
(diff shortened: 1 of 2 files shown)


Pr0methean:
This repo is no longer maintained, so I've replaced this PR with https://github.com/zip-rs/zip2/pull/96.

